PR #1, opened 2017-11-08
Initial release

delvedor:
As titled, feedbacks?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,14 +5,15 @@
   "main": "index.js",
   "dependencies": {
     "fastify-plugin": "^0.1.1",
-    "iltorb": "^2.0.2",
-    "pump": "^1.0.2",
+    "mime-db": "^1.31.0",
+    "pump": "^1.0.3",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "fastify-compress",
+  "version": "0.1.0",
+  "description": "Fastify compression utils",
+  "main": "index.js",
+  "dependencies": {
+    "fastify-plugin": "^0.1.1",
+    "iltorb": "^2.0.2",
```
mcollina: this is a native dependency. I would not add it as a dependency here, bur rather treat it as optional (in case it fails to install).

delvedor:
> Should this be implemented with a onSend hook that alters the payload?

Yes, I'm waiting to a pr that allows me to alter the payload completely, at the moment is not possible.
Furthermore, this plugin adds the `compress` decorator.

package.json
delvedor: 👍

PR #6, opened 2017-11-19
Allow missing accept-encoding header

SerayaEryn:
According to [https://tools.ietf.org/html/rfc7231#section-5.3.4](https://tools.ietf.org/html/rfc7231#section-5.3.4) a missing accept-encoding header is valid.

In this case no error should be send and the payload should not be compressed.

### DIFF
--- a/test.js
+++ b/test.js
@@ -146,7 +146,7 @@ test('should not compress on missing header', t => {
   })
 })
 
-test('Should not compress on missing header', t => {
+test('Should close the stream', t => {
   t.plan(3)
   const fastify = Fastify()
   fastify.register(compressPlugin, { global: false })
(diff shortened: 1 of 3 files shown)


test.js
```diff
@@ -141,17 +141,12 @@ test('Missing header', t => {
...
-      message: 'Missing `accept encoding` header',
-      statusCode: 400
-    }, payload)
+    t.strictEqual(res.statusCode, 200)
+    t.notOk(res.headers['content-encoding'])
   })
 })
 
-test('Should close the stream', t => {
+test('Should not compress on missing header', t => {
```
delvedor: Can you update this test?
You can pass an invalid header, we want to check that the plugin will close the stream in case of error :)

SerayaEryn:
I changed that. :)

PR #7, opened 2017-11-20
Support asterisk in accept-encoding header

SerayaEryn:
This PR adds support for an asterisk `*` in the `accept-encoding` header.

### DIFF
--- a/index.js
+++ b/index.js
@@ -155,6 +155,9 @@ function getEncodingHeader (request) {
     if (supportedEncodings.indexOf(acceptEncodings[i]) > -1) {
       return acceptEncodings[i]
     }
+    if (acceptEncodings[i].indexOf('*') > -1) {
+      return 'gzip'
+    }
   }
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -155,6 +155,9 @@ function getEncodingHeader (request) {
     if (supportedEncodings.indexOf(acceptEncodings[i]) > -1) {
       return acceptEncodings[i]
     }
+    if (acceptEncodings[i].indexOf('*') > -1) {
```
delvedor: I think this can be faster.
If a user puts `*` as accepted encodings, there will not be other stuff there.
if so, we can just check it the header value is equal to `*`.

Does it make sense?

SerayaEryn: No, we cannot do that. Something like 'gzip;q=1.0, *;q=0.5' is valid and might occur. This would mean use gzip if possible and else use any other encoding.

PR #8, opened 2017-11-20
Increase test coverage

SerayaEryn:
As titled.

### DIFF
--- a/test.js
+++ b/test.js
@@ -179,12 +179,36 @@ test('Should close the stream', t => {
     stream.on('close', () => t.ok('stream closed'))
     reply.type('text/plain').compress(stream)
   })
+  
+    fastify.inject({
+      url: '/',
+      method: 'GET'
+    }, res => {
+      const payload = JSON.parse(res.payload)
+      t.strictEqual(res.statusCode, 400)
+      t.deepEqual({
+        error: 'Bad Request',
+        message: 'Missing `accept encoding` header',
+        statusCode: 400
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -372,6 +399,117 @@ test('Should compress json data (gzip) - global', t => {
...
+})
+
+test('Should not compress if content-type is a invalid type', t => {
+  t.plan(2)
+  const fastify = Fastify()
+  fastify.register(compressPlugin, { threshold: 0 })
+  const json = { hello: 'world' }
+
+  fastify.get('/', (req, reply) => {
+    reply.header('content-type', 'something/invalid')
```
delvedor: This test is not needed, `reply.type` is an alias of `reply.header('content-type', ` :)

test.js
```diff
@@ -372,6 +399,117 @@ test('Should compress json data (gzip) - global', t => {
...
+
+  fastify.inject({
+    url: '/',
+    method: 'GET',
+    headers: {
+      'accept-encoding': 'deflate'
+    }
+  }, res => {
+    t.strictEqual(res.statusCode, 200)
+    t.notOk(res.headers['content-encoding'])
```
delvedor: Can you check also the payload?

SerayaEryn:
I'll fix that as soon as possible.

SerayaEryn:
I fixed the tests and I am now checking the payload in the tests.

PR #17, opened 2018-01-23
Updated to support the latest version of fastify

delvedor:
As titled.
Also added support for streams 1, see https://github.com/fastify/fastify/pull/648.


delvedor:
@mcollina I used the same example you did in https://github.com/fastify/fastify/pull/648 with `jsonstream` :)

PR #32, opened 2018-08-27
make compressibleTypes user settable so custom types can be compressed

tobinbradley:
The itch I'm scratching is I couldn't compress a protobuf content type, as it wasn't in mime-db or in the `compressibleTypes` constant. I was going to add it to `compressibleTypes`, but that didn't seem like a sustainable solution, so I made the `compressibleTypes` regex user configurable.

Since the types that were in `compressibleTypes` are also in mime-db, I'm guessing this is used to short circuit the mime-db check for performance. Figured overwriting `compressibleTypes` this way would be more performant than adding an additional regex check for a user passed expression (i.e. compressibleTypes.test(type) || someuserstuff.test(type)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,14 @@ fastify.register(
   { threshold: 2048 }
 )
 ```
+### Compression Header Customization
+[mime-db](https://github.com/jshttp/mime-db) is used to determine if a `Content-Type` should be compressed. You can compress additional content types via regular expression.
+```javascript
+fastify.register(
+  require('fastify-compress'),
+  { customTypes: /^x-protobuf$/ }
+)
+```
 ### Brotli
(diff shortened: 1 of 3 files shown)


tobinbradley:
I added 3 unit tests - check to see if it compresses if custom content type matches, that it doesn't compress if custom content type doesn't match, and that it won't compress (or break) if somebody passes in not-regex. Also cleaned up the docs for that section a bit.

Kudos on your test suite. I'm new to unit testing, but the way you did it with fastify.inject was really easy to pick up.

tobinbradley:
I should also mention the unit tests for custom content type are using junk values - I didn't want to use anything that might become a real content type some day.

delvedor:
Sorry for the delay!
Landed in `v0.7.0` :)

tobinbradley:
Awesome, and thanks for the great project!

Issue #28, opened 2018-04-19
Should suppport Buffer


mcollina:
As titled.

Currently Buffer objects are rendered as JSON.

PR #35, opened 2018-09-14
Added buffer support

cemremengu:
Addresses #28

~Added a more generic `whatever-to-stream` function instead of `sts` so that buffers can also benefit.~

Switch to [into-stream](https://github.com/sindresorhus/into-stream). If you are not comfortable with it I can just use [this](https://github.com/fastify/fastify-compress/blob/47d89dc9d2bd3c0463a86f89fafffbeab5b75297/streamify.js) again


trxcllnt:
I think I just added this to #34 without knowing there was already an outstanding PR for it

cemremengu:
I will close this one then if yours is merged @trxcllnt

I think it is better if they are separate though

PR #39, opened 2018-10-23
Remove "missing payload" warning in onSend hook

### LINKED ISSUE #37: About warning compress: missing payload'
In global mode, I'm getting "compress: missing payload" when I use `reply.redirect`.

Logging this warning directly in the `onSend` global hook when there's no payload seems a bit too much
https://github.com/fastify/fastify-compress/blob/c6afb98d04980254e700d23eaa5561e29f377593/index.js#L79-L82

Wykks:
In onSend hook, there's no reason to warn when there's no payload. It's valid to send stuff without payload, like a redirect, even when using fastify-compress in global mode.

close #37

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,7 +77,6 @@ function compressPlugin (fastify, opts, next) {
 
   function onSend (req, reply, payload, next) {
     if (payload == null) {
-      reply.res.log.warn('compress: missing payload')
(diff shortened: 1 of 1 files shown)


mcollina:
I'm good either way.

PR #48, opened 2019-03-06
feat: support native Brotli

### LINKED ISSUE #45: Native Brotli implementation
Hey :wave: 

As Brotli is natively available in Node 11, could we use that (if the used Node version supports it) by default? :thinking:

clarkdo:
Resolve #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,10 @@ fastify.register(
 )
 ```
 ### Brotli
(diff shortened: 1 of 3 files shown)


Acanguven:
Any update on this?

Eomm:
@Acanguven thank you for pop up this PR, we will take care soon

@mcollina, since the default behaviour is changed should be this change major though is the http-header to choose the accepted compressions?

mcollina:
I don’t think so, this is a minor change.

Issue #61, opened 2019-05-29
Brotli is never chosen


TJKoury:
Current implementation looks for Accepted-Encoding headers, which are sent in arbitrary order by most browsers, and[ returns the first match](https://github.com/fastify/fastify-compress/blob/master/index.js#L168).  

Plugin needs to have a preferred order option.

kumarabhishek:
@TJKoury , @mcollina  Reason for `brotli` never being chosen is `encoding-negotiator` always returning 'gzip'.  I'm using following workaround:
```
function getEncodingHeader (supportedEncodings, request) {
  const encodingNegotiator = new Negotiator(request);
  // Trying first with brotli
  if(supportedEncodings.includes('br')) {
      return 'br';
  } else {
    return encodingNegotiator.encoding(supportedEncodings)    
  }
}
```

Eomm:
> Current implementation looks for Accepted-Encoding headers, which are sent in arbitrary order by most browsers, and[ returns the first match](https://github.com/fastify/fastify-compress/blob/master/index.js#L168).
> 
> Plugin needs to have a preferred order option.

Could you elaborate?
Do you have a header example and your plugin setup? Do you set the `{ brotli: require('iltorb') }` parameter?

kumarabhishek:
@Eomm  I am using Node.js 12.9.1, so setting `{ brotli: require('iltorb') }` is not needed. I guess for default header in case of chrome `Accept-Encoding: gzip, deflate, br`, it should return 'br'. Atleast i want to prefer 'br' first and then gzip as fallback.

Eomm:
Clear, right now to archive this you could:
- add a `preHandler` that change the header removing all the encoding if `br` is in
- update the client to add the quality score in the header

There is an open PR for this feature but it is stale, would you like to work on it?

mcollina:
I would say a PR on this would be highly welcomed.

PR #71, opened 2019-08-03
Add support for quality values

### LINKED ISSUE #70: Bug: Quality value syntax not supported
It seems like fastify-compress does not support [Quality Value](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding#Syntax) syntax in the `Accepts-Encoding` header. This values for example causes a `406 Unsupported Encoding` HTTP error to be returned. 

Looking int the source of [`index.js:128`](https://github.com/fastify/fastify-compress/blob/master/index.js#L126) in `onSend` the `encoding` variable is indeed `null`. This error goes away once removing the `q` directives from the header value.

Example `Accepts-Encoding` header value that we fail to parse:
`gzip;q=1.0,deflate;q=0.6,identity;q=0.3`

Quickly looking at the source, we will likely need to update`getEncodingHeader` to support this.

SerayaEryn:
This adds support for quality values in the `accept-encoding` header.

Closes #70

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,6 +13,7 @@ const isGzip = require('is-gzip')
 const isZip = require('is-zip')
 const unZipper = require('unzipper')
 const isDeflate = require('is-deflate')
+const encodingNegotiator = require('encoding-negotiator')
 
 function compressPlugin (fastify, opts, next) {
   fastify.decorateReply('compress', compress)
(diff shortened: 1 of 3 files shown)


mcollina:
Would this help with https://github.com/fastify/fastify-compress/issues/61?

SerayaEryn:
> Would this help with #61?

It will help if Brotli has the highest quality value. 

The encodings in the `accept-encoding` header will be sorted by their quality value and the encodings with the higher quality values will be considered first. 
If Brotli has the highest quality value it will be chosen.

SerayaEryn:
@delvedor It would be nice if you could take a look at this PR.

mcollina:
@delvedor is on vacation!

tigt:
On Firefox 70, unfortunately no quality values are sent in the request header:

```
Accept-Encoding: gzip, deflate, br
```

So this alone is not sufficient for #61.

mcollina:
@tigt a PR would be nice

PR #84, opened 2019-10-22
Allow customizing used encodings

### LINKED ISSUE #43: Is it possible to force only Brotli?
I'm creating a custom server just for myself and I want it to only compress using Brotli.

### LINKED ISSUE #61: Brotli is never chosen

tigt:
While implementing the option to choose the order of accepted encodings, it was less code and more elegant to also let consumers avoid certain encodings entirely. This would be useful in the case of misbehaving downstream boxes.

Also changes default priority to prefer Brotli to gzip.

Fixes #61, fixes #43, closes #62.

## Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -153,7 +153,7 @@ function onEnd (err) {
   if (err) this.res.log.error(err)
 }
 
-function getEncodingHeader (supportedEncodings, request) {
+function getEncodingHeader (encodings, request) {
   let header = request.headers['accept-encoding']
   if (header != null) {
     header = header.toLowerCase()
--- a/index.js
+++ b/index.js
@@ -37,11 +37,17 @@ function compressPlugin (fastify, opts, next) {
   const supportedEncodings = ['gzip', 'deflate', 'identity']
   if (opts.brotli) {
     compressStream.br = opts.brotli.compressStream
-    supportedEncodings.push('br')
+    supportedEncodings.unshift('br')
   } else if (zlib.createBrotliCompress) {
     compressStream.br = zlib.createBrotliCompress
-    supportedEncodings.push('br')
+    supportedEncodings.unshift('br')
   }
+
+  const encodings = Array.isArray(opts.encodings)
+    ? supportedEncodings
+      .filter(encoding => opts.encodings.includes(encoding))
+      .sort((a, b) => opts.encodings.indexOf(a) - supportedEncodings.indexOf(b))
+    : supportedEncodings
 
   next()
 
@@ -56,10 +62,10 @@ function compressPlugin (fastify, opts, next) {
     var noCompress =
       // don't compress on x-no-compression header
       (this.request.headers['x-no-compression'] !== undefined) ||
-      // don't compress if not one of the indiated compressible types
+      // don't compress if not one of the indicated compressible types
       (shouldCompress(this.getHeader('Content-Type') || 'application/json', compressibleTypes) === false) ||
       // don't compress on missing or identity `accept-encoding` header
-      ((encoding = getEncodingHeader(supportedEncodings, this.request)) === undefined || encoding === 'identity')
+      ((encoding = getEncodingHeader(encodings, this.request)) === undefined || encoding === 'identity')
 
     if (noCompress) {
       if (inflateIfDeflated && isStream(stream = maybeUnzip(payload, this.serialize.bind(this)))) {
@@ -112,7 +118,7 @@ function compressPlugin (fastify, opts, next) {
       // don't compress if not one of the indiated compressible types
       (shouldCompress(reply.getHeader('Content-Type') || 'application/json', compressibleTypes) === false) ||
       // don't compress on missing or identity `accept-encoding` header
-      ((encoding = getEncodingHeader(supportedEncodings, req)) === undefined || encoding === 'identity')
+      ((encoding = getEncodingHeader(encodings, req)) === undefined || encoding === 'identity')
 
     if (noCompress) {
       if (inflateIfDeflated && isStream(stream = maybeUnzip(payload))) {
@@ -162,9 +168,9 @@ function closeStream (payload) {
   }
 }
 
-function getEncodingHeader (supportedEncodings, request) {
-  const header = request.headers['accept-encoding']
-  return encodingNegotiator.negotiate(header, supportedEncodings)
+function getEncodingHeader (encodings, request) {
+  const header = (request.headers['accept-encoding'] || '').replace('*', 'gzip')
+  return encodingNegotiator.negotiate(header, encodings)
 }
 
 function shouldCompress (type, compressibleTypes) {
--- a/index.js
+++ b/index.js
@@ -44,7 +44,7 @@ function compressPlugin (fastify, opts, next) {
   }
 
   if (opts.encodings && opts.encodings.length < 1) {
-    throw Error('The `encodings` array option must have at least 1 item.')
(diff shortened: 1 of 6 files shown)


tigt:
If this approach works for you, I’ll add tests to finish the MR.

index.js
```diff
@@ -37,12 +37,18 @@ function compressPlugin (fastify, opts, next) {
...
     compressStream.br = zlib.createBrotliCompress
-    supportedEncodings.push('br')
+    supportedEncodings.unshift('br')
   }
 
+  const encodings = Array.isArray(opts.encodings)
+    ? supportedEncodings
+      .filter(encoding => opts.encodings.includes(encoding))
+      .sort((a, b) => opts.encodings.indexOf(a) - supportedEncodings.indexOf(b))
+    : supportedEncodings
```
mcollina: I think we should throw if this array is empty

index.js
```diff
@@ -162,9 +168,9 @@ function closeStream (payload) {
   }
 }
 
-function getEncodingHeader (supportedEncodings, request) {
-  const header = request.headers['accept-encoding']
-  return encodingNegotiator.negotiate(header, supportedEncodings)
+function getEncodingHeader (encodings, request) {
+  const header = (request.headers['accept-encoding'] || '').replace('*', 'gzip')
```
mcollina: can you please explain this?

mcollina:
Go ahead!

index.js
tigt: Ah, yeah. That was to preserve the existing behavior where a client accepting `*` was interpreted as `gzip` — otherwise it sends Brotli with this patch. I wasn’t sure whether that was behavior I should preserve or not.

mcollina: Great, would you mind adding a comment?

SerayaEryn: I think it would be better to to change line 40 & line 43 back to use push and remove this.
It chooses the first encoding of the encoding list if a client sends `*`.

tigt: If 40/43 use `.push`, #61 happens. I’m not against your suggestion, but it seemed odd for the compression plugin for a server ecosystem focused on speed to not use the best encoding by default.

index.js
tigt: I’m having trouble figuring out how I should test this. This doesn’t pass, for example:

```js
test('Should throw if `encodings` array is empty', t => {
  const fastify = Fastify()
  t.throws(
    fastify.register(compressPlugin, { encodings: [] }),
    {}
  )
})
```

tigt:
I copied [the approach `point-of-view` is using](https://github.com/fastify/point-of-view/blob/9f9cca71701bd5567590ca7599c16d3e2bb366dd/test/test.js#L91), but still no dice:

```js
test('Should error if `encodings` array is empty', t => {
  const fastify = Fastify()
  fastify.register(compressPlugin, { encodings: [] })
  fastify.ready(err => { t.ok(err instanceof Error) })
})
```

index.js
mcollina: if encoding is empty, calll the callback with an error.

index.js
SerayaEryn: I am fine with changing the preferred encoding to `br`.
But if we do so we should also return `br` for `*` and change the test accordingly.

index.js
SerayaEryn: `t.throws` expects a function as the first parameter: https://node-tap.org/docs/api/asserts/#tthrowsfn-expectederror-message-extra

```js
test('Should throw if `encodings` array is empty', t => {
  const fastify = Fastify()
  t.throws(() => {
    fastify.register(compressPlugin, { encodings: [] })
  })
})
```

index.js
mcollina: I’m not, gzip has been widely supported for a decade at least, while brotli is a new thing. I would like to be conservative. I’m happy to have an option to change this.

tigt: That was my thinking as well. I’m not against a `bleedingEdgeEncodingAssumption` (or whatever), but I’m not certain the benefit outweighs the possible breakage and added one more source of configuration entropy.

tigt:
@mcollina, @SerayaEryn : I managed to get the tests working thanks to your help. Would you mind giving this the final lookover to see if it’s mergeworthy?

index.js
```diff
@@ -37,10 +37,24 @@ function compressPlugin (fastify, opts, next) {
...
+
+  const encodings = Array.isArray(opts.encodings)
+    ? supportedEncodings
+      .filter(encoding => opts.encodings.includes(encoding))
+      .sort((a, b) => opts.encodings.indexOf(a) - supportedEncodings.indexOf(b))
+    : supportedEncodings
+
+  if (encodings.length < 1) {
+    next(new Error('None of the passed `encodings` were supported — compression not possible.'))
   }
```
mcollina: Can you add a test that triggers this? Passing an not existing encoding would do the trick

index.js
SerayaEryn: > I’m not, gzip has been widely supported for a decade at least, while brotli is a new thing. I would like to be conservative. I’m happy to have an option to change this.

That makes sense to me.

> If 40/43 use `.push`, #61 happens. I’m not against your suggestion, but it seemed odd for the compression plugin for a server ecosystem focused on speed to not use the best encoding by default.

Might you please add a test with `Accept-Encoding: gzip, deflate, br` as mentioned in #61. This test would still return `gzip`. 
After thinking a bit about this I noticed that the problem from #61 remains with the mentioned header since the `encoding-negotiator` module sorts the encodings from the header by the quality values, which in this case are all 1 and then returns the first matching supported encoding.
I wrote a (failing) test to show that: https://github.com/SerayaEryn/encoding-negotiator/pull/10/files
Because of that a change to the `encoding-negotiator` module is required. I can do that tomorrow.

The changes to the lines 42, 43 & 172 however are not necessary.

SerayaEryn: I opened a PR in the `encoding-negotiator` module, that adds a `prefferedEncoding` option: https://github.com/SerayaEryn/encoding-negotiator/pull/10
Please let me know if this solution looks good to you.

tigt: I wonder if an array of encodings in preference order would be more robust (e.g. you ideally want Brotli, but maybe the Node you’re running on doesn’t support it. Then, it would fall back to the next encoding in the array.)

The other nit is I believe it [should be spelled](https://en.wiktionary.org/wiki/preferred) `preferredEncoding`.

mcollina:
Would you mind to rebase?

tigt:
Should I wait for https://github.com/SerayaEryn/encoding-negotiator/pull/10 before rebasing?

Also, this is kind of silly, but I’m very bad at using Git. I can eventually figure it out, but if you prefer a cleaner history for PRs, do you think [GitHub’s PR squashing feature](https://github.blog/2016-04-01-squash-your-commits/) would be a good fit for this repo?

mcollina:
The issue is that there are conflicts against master. You can also merge master if you prefer.

tigt:
Ah, beans, you’re right. I’ll try and figure this out.

tigt:
I resolved the conflicts with the GitHub UI; sorry if that’s embarrassing, but I _really_ have a hard time with Git.

tigt:
@delvedor: Do I implement the version bump and publish, or is that machinery external to the PR system?

TJKoury:
Nice, thanks @tigt for picking this up.

PR #86, opened 2019-10-25
Remove HTTP406 errors, allow variable case in accept encoding header to align with RFC 7231

### LINKED ISSUE #85: Not compliant with RFC 7231
## 🐛 Bug Report

https://tools.ietf.org/html/rfc7231#section-5.3.4

- "If an Accept-Encoding header field is present in a request
   and none of the available representations for the response have a
   content-coding that is listed as acceptable, the origin server SHOULD
   send a response without any content-coding."
- "All content-coding values are case-insensitive"

## To Reproduce

- A request with the header `Accept-Encoding: GzIP` responds HTTP406.
- A request with the header `Accept-Encoding: Invalid` responds HTTP406.

## Expected behavior

- A request with the header `Accept-Encoding: GzIP` should respond with gzip compressed payload.
- A request with the header `Accept-Encoding: Invalid` should respond with uncompressed payload.

## Your Environment

- *node version*: v12.13.0
- *fastify version*: 2.10.0
- *os*: Mac

j-windsor:
Fixes #85 

#### Checklist

- [Ran test, benchmark is not a script] run `npm run test` and `npm run benchmark`
- [Tests succeeded] tests and/or benchmarks are included
- [Yes] documentation is changed or added
- [YES] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,7 +59,7 @@ function compressPlugin (fastify, opts, next) {
       // don't compress if not one of the indiated compressible types
       (shouldCompress(this.getHeader('Content-Type') || 'application/json', compressibleTypes) === false) ||
       // don't compress on missing or identity `accept-encoding` header
-      ((encoding = getEncodingHeader(supportedEncodings, this.request)) === undefined || encoding === 'identity')
+      ((encoding = getEncodingHeader(supportedEncodings, this.request)) == null || encoding === 'identity')
 
     if (noCompress) {
       if (inflateIfDeflated && isStream(stream = maybeUnzip(payload, this.serialize.bind(this)))) {
@@ -69,12 +69,6 @@ function compressPlugin (fastify, opts, next) {
         pump(stream, payload = unzipStream(uncompressStream), onEnd.bind(this))
       }
       return this.send(payload)
-    }
-
-    if (encoding === null) {
-      closeStream(payload)
-      this.code(406).send(new Error('Unsupported encoding'))
-      return
     }
 
     if (typeof payload.pipe !== 'function') {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -71,12 +71,6 @@ function compressPlugin (fastify, opts, next) {
       return this.send(payload)
     }
 
-    if (encoding === null) {
```
Eomm: I would leave this feature as opt-in since it could be willing for the user to choose what to do since the RFC say "should" and not "must".

We should choose also the default in order to decide if this fix will be semver-major or not

mcollina:
@delvedor do you recall why you added the 406 error in https://github.com/fastify/fastify-compress/commit/d3c322a9b07f6e64b1378edbd369e8fc56604d69?

I think we should bump the major of this module anyway.

index.js
mcollina: I’ve read the spec, and I do not see a good reason for the old behavior to be there.

Eomm: The reason would be to apply the 406 logic also to the encoding and be apply strict validation to the requests like the other headers. The [`should` definition](https://tools.ietf.org/html/rfc2119) may consent this behaviour that we are removing. _To be even more open this logic could be an optional callback with the 406 logic_

Of course, this must be interpreted as an additional feature not a standard in any way.

j-windsor: I agree that it would be nice to allow the user to choose how they want to deal with the case where the encoding is invalid, but in no way is it required, and I can't imagine too many people would want to override standard behavior that we *should* have (defined by IETF).

I could see having an optional callback in the plugin options and the compress function arguments. @mcollina what do you think of this?

j-windsor:
Feel free to merge. I will follow up with custom handling. should it be implemented as an optional callback?

delvedor:
Yes, offer it as an option of the plugin, for example:
```js
fastify.register(require('fastify-compress'), {
  onUnsupportedEncoding: (encoding, req, reply) => {
    reply.code(406).send(new Error(`Unsupported encoding: ${encoding}`))
  }
})
```

delvedor:
I'll wait for the pr before to cut the release :)
Thank you again!

Issue #89, opened 2019-11-05
How to compress the output of an ejs template rendered reply without set the global flag


red1:
Hello 
i'm doing something like that : 
`reply.view('/src/templates/hello.ejs', { hello: 'world' })
`
I'd like to compress the output of my ejs template, .compress will not work directly and I don't want to activate the global flag
Do you know how can I do that ? can we enrich the component with a compressOnSend method to accept to be chained and activate the onsend hook ?
`reply.view('/src/templates/hello.ejs', { hello: 'world' }).compressOnSend()`
Thanks

mcollina:
I think this is better solved in the fastify-compress module, adding support for compressing a single route. This is already possible given the internals of the framework, we are only missing support in fastify-compress.

red1:
It would be great yes please!

red1:
Hello @mcollina 
Do you know if this support will be added ?

mcollina:
If somebody would like to work on it, yes!

Eomm:
You could try to implement it 💪

this plugin: https://github.com/fastify/fastify-rate-limit has this kind of feature: global and/or per-route usage, you may get some inspiration from it!

marc-fisher-q2e:
Is there a way to hook into onSend per route? The last per route option in the lifecycle right now is preSerialization so any thoughts on how this could work?

Eomm:
> Is there a way to hook into onSend per route?

No, right now there is not this option, but it should easy to add here:
https://github.com/fastify/fastify/blob/master/lib/route.js#L9

> so any thoughts on how this could work?

Good point, we should add to fastify that parameter, then this issue can be implemented

Svjard:
https://github.com/fastify/fastify-compress/pull/92

PR #92, opened 2019-12-07
Feature\#89 - Per route options for compression

Svjard:
#### Checklist

- [X] run `npm run test` and `npm run benchmark`
- [X] tests and/or benchmarks are included
- [X] documentation is changed or added
- [X] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)


Svjard:
Took a first pass at this welcome feedback on it. In particular:
1) Removed the `decorate` call to instead have compress() flow from the start of the request on the reply. Because each route can be configured differently was hard to determine the best approach from the previous approach.
2) Better tests for per route, have a minimal set but open to writing more if you have some good ones I can add.
3) The package.json is pointing to master, I know this needs to change but waiting for the next Fastify publish.

package.json
```diff
@@ -13,6 +13,7 @@
     "is-zip": "^1.0.0",
     "mime-db": "^1.33.0",
     "minipass": "^2.9.0",
+    "moment": "^2.24.0",
```
Eomm: Is this necessary?

test/test-routes.js
```diff
@@ -0,0 +1,129 @@
...
+      const file = readFileSync('./package.json', 'utf8')
+      const payload = zlib.inflateSync(res.rawPayload)
+      t.strictEqual(payload.toString('utf-8'), file)
+      t.strictEqual(usedCustom, true)
+      t.strictEqual(usedCustomGlobal, false)
+    })
+  })
+})
+
+test('should send a gzipped data with custom zlib', t => {
```
Eomm: I would add a test that checks that the compression is not applied when global false and there is a "compressed route"

package.json
```diff
@@ -23,7 +24,7 @@
     "@types/node": "^12.0.8",
     "@typescript-eslint/parser": "^2.0.0",
     "eslint-plugin-typescript": "^0.14.0",
-    "fastify": "^2.0.0",
+    "fastify": "fastify/fastify#master",
```
Eomm: It should be released in 2.11 👍

index.js
```diff
@@ -16,159 +16,221 @@ const isDeflate = require('is-deflate')
...
+
+  // add onSend hook onto each route as needed
+  fastify.addHook('onRoute', (routeOptions) => {
+    if (routeOptions.config && typeof routeOptions.config.compress !== 'undefined') {
+      if (typeof routeOptions.config.compress === 'object') {
+        const mergedCompressParams = Object.assign({}, globalParams, processParams(routeOptions.config.compress))
+        // if the current endpoint has a custom compress configuration ...
+        buildRouteCompress(fastify, mergedCompressParams, routeOptions)
+      } else if (routeOptions.config.compress === false) {
+        // don't apply any compress settings
```
Eomm: this case is not documented

index.js
```diff
@@ -16,159 +16,221 @@ const isDeflate = require('is-deflate')
...
+      } else {
+        throw new Error('Unknown value for route compress configuration')
```
Eomm: we should add a test for this case as well

index.js
```diff
@@ -16,159 +16,221 @@ const isDeflate = require('is-deflate')
...
     }
 
     var stream, encoding
     var noCompress =
       // don't compress on x-no-compression header
-      (this.request.headers['x-no-compression'] !== undefined) ||
-      // don't compress if not one of the indicated compressible types
-      (shouldCompress(this.getHeader('Content-Type') || 'application/json', compressibleTypes) === false) ||
+      (req.headers['x-no-compression'] !== undefined) ||
+      // don't compress if not one of the indiated compressible types
```
Eomm:
```suggestion
      // don't compress if not one of the indicated compressible types
```

index.js
```diff
@@ -16,159 +16,221 @@ const isDeflate = require('is-deflate')
 const encodingNegotiator = require('encoding-negotiator')
 
 function compressPlugin (fastify, opts, next) {
-  fastify.decorateReply('compress', compress)
```
mcollina: Can you set this to `null` instead? In this way it will be added to the prototype

index.js
```diff
@@ -16,159 +16,221 @@ const isDeflate = require('is-deflate')
...
-
-  const encodings = Array.isArray(opts.encodings)
+  params.encodings = Array.isArray(opts.encodings)
     ? supportedEncodings
       .filter(encoding => opts.encodings.includes(encoding))
       .sort((a, b) => opts.encodings.indexOf(a) - supportedEncodings.indexOf(b))
     : supportedEncodings
 
-  if (encodings.length < 1) {
-    next(new Error('None of the passed `encodings` were supported — compression not possible.'))
```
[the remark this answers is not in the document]
Svjard: It still exists above, we just build out the parameters first in processParams() and then the two checks with the 2 next() statements follow.

index.js
```diff
@@ -16,159 +16,221 @@ const isDeflate = require('is-deflate')
...
+  const globalParams = processParams(opts)
 
-  if (opts.global !== false) {
-    fastify.addHook('onSend', onSend)
+  if (opts.encodings && opts.encodings.length < 1) {
+    next(new Error('The `encodings` option array must have at least 1 item.'))
+  }
+
+  if (globalParams.encodings.length < 1) {
+    next(new Error('None of the passed `encodings` were supported — compression not possible.'))
```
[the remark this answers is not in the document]
Svjard: yep, thanks fixed.

mcollina:
https://github.com/fastify/fastify-compress/pull/92/files#diff-168726dbe96b3ce427e7fedce31bb0bcR314

Can you bump this to fastify v2.11? That's the one you'll need for this fix.

mcollina:
(this will be a semver-major change).

test/test-routes.js
marc-fisher-q2e: Test added.

package.json
marc-fisher-q2e: Done, thanks

index.js
marc-fisher-q2e: Added `Note` in the README about being able to exclude compression on a route with this.

index.js
marc-fisher-q2e: Are you saying test the throw condition? Or the compress=false condition above it?

index.js
marc-fisher-q2e: Thanks, added it back in and set it to null so added on the prototype later on the onrequest hook

index.js
Eomm: Yes, I mean the throw 👍

Svjard: Added

index.js
```diff
@@ -16,159 +16,225 @@ const isDeflate = require('is-deflate')
...
+    routeOptions.onRequest.push(onRequest)
+  } else if (typeof routeOptions.onRequest === 'function') {
+    routeOptions.onRequest = [routeOptions.onRequest, onRequest]
+  } else {
+    routeOptions.onRequest = [onRequest]
   }
 
-  next()
+  function onRequest (req, reply, next) {
+    reply.compress = compress(params)
```
mcollina: Given that `params` will not change, can you cache the output of `compress(params)` outside of the `onRequest` hook?

Svjard: 👍

PR #102, opened 2020-02-14
bug fix - undefined value of compressible now does not compress

adityapatadia:
https://github.com/jshttp/mime-db package has many content type for which the value of `compressible` is undefined. When this is encountered, this package was wrongly compressing those mime types. One such example is `image/webp` type.

This PR fixes that wrong behavior and also adds test to prevent it from happening again in future.

#### Checklist


adityapatadia:
Thanks for quick merge. A release would also be great so we can use this in production.

mcollina:
it's already out.

adityapatadia:
Awesome! Thanks a ton.

PR #107, opened 2020-04-23
V3 upgrade

thomheymann:
https://github.com/fastify/fastify/issues/2217

#### Checklist


package.json
```diff
@@ -23,19 +23,20 @@
     "@types/node": "^13.1.0",
     "@typescript-eslint/parser": "^2.19.2",
     "eslint-plugin-typescript": "^0.14.0",
-    "fastify": "^2.11.0",
-    "iltorb": "2.4.3",
+    "fastify": "^3.0.0-alpha.1",
```
jsumners, line 26: What is the plan here, @mcollina? Are we going to release all of the core modules targeting the alpha releases and then come back and do _another_ pass to update them to the final v3?

mcollina: What would you prefer? I think it's the best path forward. Hopefully if dependabot works we can automate some of it.

package.json
```diff
@@ -23,19 +23,20 @@
...
+    "iltorb": "^2.4.5",
```
mcollina, line 27: Can you remove this? I _think_ node 10 has native brotli.

index.d.ts
```diff
@@ -1,27 +1,25 @@
-import { Plugin, FastifyReply, FastifyRequest } from 'fastify'
-import { Server, IncomingMessage, ServerResponse } from 'http'
+import { FastifyPlugin, FastifyReply, FastifyRequest, RawServerBase } from 'fastify'
 import { Stream } from 'stream';
+import { Input, InputObject } from 'into-stream';
```
mcollina, line 3: I would prefer we did not expose this details to the users.

package.json
jsumners: Sounds good. Let's learn how well dependabot helps out with this.

index.d.ts
mcollina: this is ok.

index.js
```diff
@@ -314,6 +311,6 @@ function unzipStream (inflate, maxRecursion) {
 }
 
 module.exports = fp(compressPlugin, {
-  fastify: '>=2.11.0',
+  fastify: '>=3',
```
mcollina, line 314:
```suggestion
  fastify: '3.x',
```

Issue #105, opened 2020-03-24
Should fastify-compress handle compressed income requests?


joelwallis:
## 🚀 Feature Proposal

A clear and concise description of what the feature is.

Fastify currently does not accept requests with `Content-Encoding: gzip` in its normal form, and since there is a generic module for compression-related work in the Fastify ecosystem, I think it would be good to extend its capabilities to support compressed incoming requests.

## Motivation

Compression in HTTP is useful in HTTP requests too, not only in HTTP responses.

## Example

An HTTP client would send a compressed HTTP request to a Fastify server using fastify-compress, and it would be handled properly.

mcollina:
I agree, it would be a good feature! Would you like to send a PR?

joelwallis:
I would love to! However, I don't have much experience with Fastify or its codebase, so I would happily receive guidance/feedback on my PR. :-)

mcollina:
This needs to be investigated, I'm not necessarily sure on how and it might require some effort on Fastify itself.

ShogunPanda:
Shall this be closed?

PR #110, opened 2020-05-16
Uncompress request body

ShogunPanda:
This PR adds a mechanism for fastify-compress to handle compressed request payloads.

This is achieved via the new `preDecoding` hook which is still pending in fastify/fastify#2278. If that PR is changed, this PR will be updated accordingly.

Several new options are available, please see README.md changes for more details.

Before merging this PR, we have to update the `package.json` with the right fastift dependency.

Fixes fastify/fastify#1986 and fastify/fastify-compress#105.

Hope this helps!
  Paolo

#### Checklist


index.js
```diff
@@ -79,6 +124,8 @@ function processParams (opts) {
   }
 
   const supportedEncodings = ['gzip', 'deflate', 'identity']
+
+  /* istanbul ignore next */
```
mcollina, line 128: I think we can remove the if completely right? we have brotli everywhere?

index.js
```diff
@@ -93,6 +140,51 @@ function processParams (opts) {
...
+      gzip: customZlib.createGunzip || zlib.createGunzip,
+      deflate: customZlib.createInflate || zlib.createInflate
+    },
+    encodings: [],
+    forceEncoding: null
+  }
+
+  const supportedEncodings = ['gzip', 'deflate', 'identity']
+
+  /* istanbul ignore next */
```
mcollina, line 165: same

index.js
```diff
@@ -174,6 +266,68 @@ function buildRouteCompress (fastify, params, routeOptions, decorateOnly) {
...
+
+    // Check that encoding is supported
+    if (!params.encodings.includes(encoding)) {
+      let errorPayload
+
+      if (params.onUnsupportedRequestEncoding) {
+        try {
+          errorPayload = params.onUnsupportedRequestEncoding(encoding, request)
+        } catch (ex) {
+          /* istanbul ignore next */
```
mcollina, line 300: Please add a test to cover this case.

index.js
```diff
@@ -240,6 +394,37 @@ function onEnd (err) {
...
+function onDecompressError (request, params, encoding, error) {
+  this.log.debug(`compress: invalid request payload - ${error}`)
+
+  let errorPayload
+
+  if (params.onInvalidRequestPayload) {
+    try {
+      errorPayload = params.onInvalidRequestPayload(encoding, request, error)
+    } catch (ex) {
+      /* istanbul ignore next */
```
mcollina, line 415: please add a test to cover this.

index.js
ShogunPanda: Will do

index.js
ShogunPanda: Will do

index.js
ShogunPanda: AFAIK no. fastify-compress declares support for Node >= 10.16, while brotli (according to docs) landed in 11.7.0. Am I wrong?

mcollina: It got backported in Node 10.16. See https://nodejs.org/docs/latest-v10.x/api/zlib.html

ShogunPanda: Oh, nice. I'll remove from both compress and uncompress part then.

index.js
ShogunPanda: Done

index.js
ShogunPanda: Done

index.js
ShogunPanda: Done

index.js
ShogunPanda: Done

PR #120, opened 2020-08-06
When compressing add `accept-encoding` to the Vary header, if the header exists concatenate it.

### LINKED ISSUE #108: Add Vary header to response
## 🚀 Feature Proposal

Ideally, `vary: accept-encoding` header should be added to responses which are compressed by this module. If any `vary` header is already present in response, `accept-encoding` should be added as comma separated value.

## Motivation

https://tools.ietf.org/html/rfc7234#section-4.1 and https://tools.ietf.org/html/rfc7231#section-7.1.4

Arnovsky:
closes #108 

#### Checklist


test/test-global-compress.js
```diff
@@ -29,6 +29,7 @@ test('should send a deflated data', t => {
   }, (err, res) => {
     t.error(err)
     t.strictEqual(res.headers['content-encoding'], 'deflate')
+    t.strictEqual(res.headers.vary, 'accept-encoding')
```
Arnovsky, line 32: I have added this here, since I don't think it deserves its own test, since that test will mostly be duplicated code.

index.js
```diff
@@ -371,6 +371,15 @@ function compress (params) {
       payload = intoStream(payload)
     }
 
+    if (this.hasHeader('Vary')) {
+      const varyHeader = Array.isArray(this.getHeader('Vary')) ? this.getHeader('Vary') : [this.getHeader('Vary')]
```
shalomscott, line 375: .includes('accept-encoding')

index.js
```diff
@@ -371,6 +371,15 @@ function compress (params) {
...
+      if (!varyHeader.some(h => h === 'accept-encoding')) {
```
shalomscott, line 376: `.includes('accept-encoding')`

mcollina: pr?

PR #135, opened 2020-11-01
Added Vary header to all responses

catalint:
[bug] While replying with brotly encoding, vary header was missing

Since the response body differs based on the Accept-Encoding from the request header, we need to add Vary Header of "accept-encoding" to **all** responses processed with the fastify-compress library so that CDNs and other caches work properly.

#### Checklist


mcollina (approved):
lgtm

PR #141, opened 2020-12-21
Do not set the compress onSend hook twice

### LINKED ISSUE #140: Corrupted reply with async prefixed routes and br encoding
## 🐛 Bug Report

The JSON response becomes corrupted when using br compression as well as prefixed async routes, This may be the unresolved cause of https://github.com/fastify/fastify-compress/issues/104

## To Reproduce

```js
// package.json
{
  "name": "node_test",
  "version": "1.0.0",
  "main": "index.js",
  "scripts": {
    "start": "node index.js"
  },
  "dependencies": {
    "fastify": "^3.9.2",
    "fastify-compress": "^3.4.0"
  }
}
```
```js
// index.js
const server = require('fastify')();

server.register(require('fastify-compress'), {
    threshold: 0,
    // the issue disappears by excluding 'br' from the encodings
    // encodings: ['gzip', 'deflate', 'identity']
});

server.register(require('./routes'));

server.listen(8080);
```
```js
// routes/index.js
module.exports = async (fastify, _) => {
    fastify.register(require('./test'), {
        prefix: '/test', // if this prefix is removed, the issue disappears
    });
}
```
```js
// routes/test.js
module.exports = async (fastify, opts) => {
    fastify.get('/', opts, async (req, _) => {
        return { hi: true };
    });
}
```
If a GET request is made to `http://localhost:8080/test` the response body is:
```js
�{"hi":true}
```
Note the expected behavior is produced if the prefix in routes/index.js is removed, or if 'br' encoding is explicitly removed when registering fastify-compress

## Expected behavior

```js
{"hi":true}
```

## Your Environment

- *node version*: v14.15.1
- *fastify version*: >=3.9.2
- *os*: Windows 10
- *GET request was made with the chrome fetch API & postman and results were the same*

mcollina:
Fixes https://github.com/fastify/fastify-compress/issues/140

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,6 +19,8 @@ const { inherits, format } = require('util')
 const InvalidRequestEncodingError = createError('FST_CP_ERR_INVALID_CONTENT_ENCODING', 'Unsupported Content-Encoding: %s', 415)
 const InvalidRequestCompressedPayloadError = createError('FST_CP_ERR_INVALID_CONTENT', 'Could not decompress the request payload using the provided encoding', 400)
 
+const compressAdded = Symbol('fastify-compress.added')
+
 function compressPlugin (fastify, opts, next) {
   const globalCompressParams = processCompressParams(opts)
   const globalDecompressParams = processDecompressParams(opts)
@@ -179,6 +181,16 @@ function processDecompressParams (opts) {
 }
 
 function buildRouteCompress (fastify, params, routeOptions, decorateOnly) {
+  // This methods works by altering the routeOptions, it has side effects.
+  // There is the possibility that the same options are set for more than
+  // one route, so we just need to make sure that the hook is addded only
+  // once.
+  if (routeOptions[compressAdded]) {
+    return
+  }
+
+  routeOptions[compressAdded] = true
+
   // In order to provide a compress method with the same parameter set as the route itself has
   // we do the decorate the reply at the start of the request
   if (Array.isArray(routeOptions.onRequest)) {
--- a/test/test-routes-compress.js
+++ b/test/test-routes-compress.js
@@ -263,8 +263,6 @@ test('avoid double onSend', t => {
 
   server.register(compressPlugin, {
     threshold: 0
-    // the issue disappears by excluding 'br' from the encodings
-    // encodings: ['gzip', 'deflate', 'identity']
   })
 
   server.register(async function (server) {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -179,6 +181,16 @@ function processDecompressParams (opts) {
 }
 
 function buildRouteCompress (fastify, params, routeOptions, decorateOnly) {
+  // This methods works by altering the routeOptions, it has side effects.
```
jsumners, line 184: I don't understand how this is getting added twice and why the side effect is necessary. The test is only registering the plugin once. Can you provide detail?

test/test-routes-compress.js
```diff
@@ -255,3 +255,32 @@ test('should throw an error on invalid compression setting', t => {
...
+
+test('avoid double onSend', t => {
+  t.plan(2)
+
+  const server = Fastify()
+
+  server.register(compressPlugin, {
+    threshold: 0
+    // the issue disappears by excluding 'br' from the encodings
+    // encodings: ['gzip', 'deflate', 'identity']
```
Eomm, line 267:
```suggestion
```

index.js
Eomm: It should be caused by the `ignoreTrailingSlash` that register the routes `/hello` and `/hello/`

zekth: i Confirm : 
```
/test
/test/
```


```js
    function afterRouteAdded (path, notHandledErr, done) {
      const url = prefix + path
      console.log(url)
```

RafaelGSS (approved):
LGTM.

index.js
jsumners: Excellent. That reduces this to the actual issue.

jsumners (approved):
LGTM

If there's a way to do this without mutating the options object, though, I'd prefer to see that.

delvedor (approved):
LGTM

index.js
zekth: This might be good to be documented in https://www.fastify.io/docs/latest/Hooks/#onroute don't you think? Like a `Note`.

zekth (approved):
lgtm

mcollina:
> If there's a way to do this without mutating the options object, though, I'd prefer to see that.

That will require a semver-major change in Fastify - we should probably do it.

Ethan-Arrowood (approved):

PR #146, opened 2021-03-02
fix: avoid adding "accept-encoding" to vary header if already set

### LINKED ISSUE #145: Duplicated accept-encoding vary header in response
## 🐛 Bug Report

If the `vary` header has already been set as a string with multiple concatenated values, one of which is "accept-encoding", the plugin adds a second vary header with the "accept-encoding" value. 

## To Reproduce

1- register the plugin, then set the following into a route
```js
reply.header('vary', 'accept-encoding,accept-version');

```
2- Check the response

## Expected behavior

The value should not be added if already set inside the vary header.

NOTE: the plugin works correctly if the vary header is set as an array, ie. `reply.header('vary', ['accept-encoding','accept-version']);`, but I think the string version should also be handled.

## Your Environment

- *node version*: 14.15.1
- *fastify version*: 3.12.0
- *os*: Mac

marcoreni:
#### Checklist

Fixes #145 .

### DIFF
--- a/index.js
+++ b/index.js
@@ -397,7 +397,7 @@ function compress (params) {
 function setVaryHeader (reply) {
   if (reply.hasHeader('Vary')) {
     const varyHeader = Array.isArray(reply.getHeader('Vary')) ? reply.getHeader('Vary') : [reply.getHeader('Vary')]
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #166, opened 2021-05-17
fix brotli responses being double-compressed

dwickern:
Related to https://github.com/fastify/fastify-static/pull/158

When used with fastify-static's `preCompressed` option, this plugin will double-compress `.br` files.

Currently this plugin avoids double-compressing for gzip and deflate, by inspecting the magic number in few bytes of the stream. However Brotli doesn't have a magic number to look for.

https://github.com/fastify/fastify-compress/blob/714e4a91ecf743b2bd7991b8bb9aea466992a271/index.js#L491-L494
https://github.com/fastify/fastify-compress/blob/714e4a91ecf743b2bd7991b8bb9aea466992a271/index.js#L460-L465

Instead we can check for an existing `Content-Encoding` header on the reply. This has the extra advantage that we don't need to strip off the `Content-Length` header.

#### Checklist

- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/index.js
+++ b/index.js
@@ -223,6 +223,11 @@ function buildRouteCompress (fastify, params, routeOptions, decorateOnly) {
     if (payload == null) {
       return next()
     }
+    const responseEncoding = reply.getHeader('Content-Encoding')
+    if (responseEncoding && responseEncoding !== 'identity') {
+      // response is already compressed
+      return next()
+    }
     setVaryHeader(reply)
 
     let stream, encoding
--- a/test/test-global-compress.js
+++ b/test/test-global-compress.js
@@ -195,16 +195,21 @@ test('should support quality syntax', t => {
   })
 })
 
-test('onSend hook should not double-compress Stream if already zipped', t => {
-  t.plan(3)
+test('onSend hook should not double-compress Stream if already gzipped', t => {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -223,6 +223,11 @@ function buildRouteCompress (fastify, params, routeOptions, decorateOnly) {
     if (payload == null) {
       return next()
     }
+    const responseEncoding = reply.getHeader('Content-Encoding')
+    if (responseEncoding && responseEncoding !== 'identity') {
+      // response is already compressed
+      return next()
+    }
     setVaryHeader(reply)
```
mcollina, line 226: Should the new block be after setting the setVaryHeader.

dwickern: Shouldn't fastify-static be responsible for setting the `Vary` header?

mcollina: I don't understand the answer. Anyway I'm landing - if it becomes a problem I guess we'll get another PR.

mcollina:
Unfortunately I landed without the test passing. I'm going to rever the commit, would you mind sending a fresh PR?

Thanks!

PR #173, opened 2021-06-14
fix: vary header append

### LINKED ISSUE #172: Add ability to append 'vary' response header rather than create a new one
## 🚀 Feature Proposal
At present, this plugin creates a new 'vary' header, even if one already exists, leading to the vary header becoming an array internally (`vary: ["Origin", "accept-encoding"]`) and presenting as multiple headers in a response: 

```
vary: Origin
vary: accept-encoding
```

It would be good if a plugin could append an existing vary header i.e. `vary: Origin, accept-encoding`, similar to what the [fastify-cors](https://github.com/fastify/fastify-cors) plugin does, preferably with an optional boolean option in the config.

## Motivation

Whilst the above example is perfectly valid, some clients struggle with multiple headers with the same name (legacy medical systems basically).

## Example

```js
fastify.register(
  require('fastify-compress'),
  { appendVaryHeader: true }
)
```

zekth:
Fix https://github.com/fastify/fastify-compress/issues/172

To fit the RFC: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -403,7 +403,7 @@ function setVaryHeader (reply) {
   if (reply.hasHeader('Vary')) {
     const varyHeader = Array.isArray(reply.getHeader('Vary')) ? reply.getHeader('Vary') : [reply.getHeader('Vary')]
     if (!varyHeader.some((h) => h.includes('accept-encoding'))) {
-      reply.header('Vary', [...varyHeader, 'accept-encoding'])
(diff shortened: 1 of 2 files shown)


jsumners (approved):
Looks good to me.

mcollina (approved):
lgtm

could somebody issue a release with this fix?

Issue #198, opened 2021-11-26
feat: move to `compress` and `decompress` route specific options, improved typescript types and tests, reorganize tests


darkgl0w:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Hello,

following the PR https://github.com/fastify/fastify-compress/pull/196, I would like to work on a few things for this package that will help improve development experience:

- [x] Add `compress` and `decompress` route shorthand configuration properties
- [x] Improve typescript types and actually test those types
- [x] Update the documentation
- [x] Reorganize and/or rewrite tests (renaming files following `xxx.test.js` convention, making groups with sub-tests, ...)

**Note:** some of those changes are breaking changes and will require a major bump.

If you think my proposal is acceptable the first 3 points are covered by this PR https://github.com/fastify/fastify-compress/pull/199.

Regards.

### Motivation

- It will follow the way over official packages define specific route options, it will be shorter ( :rofl: ) and **more intuitive**.
- It will have complete typings and those types will actually be tested.

### Example

Current way to configure specific route options:
```js
'use strict'

const zlib = require('zlib')
const fastify = require('fastify')()

fastify.register(require('fastify-compress', { global: false })

fastify.get('/one', {
  config: {
    compress: { createGzip: () => zlib.createGzip() },
    decompress: { createGunzip: () => zlib.createGunzip() }
  }
}, (request, reply) => {
 // ...
})

fastify.route({
  method: 'GET',
  path: '/two',
  config: {
    compress: { createGzip: () => zlib.createGzip() },
    decompress: { createGunzip: () => zlib.createGunzip() }
  },
  handler: (request, reply) => {
   // ...
  }
})
```

New way to configure specific route options:
```js
'use strict'

const zlib = require('zlib')
const fastify = require('fastify')()

fastify.register(require('fastify-compress', { global: false })

fastify.get('/one', {
  compress: { createGzip: () => zlib.createGzip() },
  decompress: { createGunzip: () => zlib.createGunzip() }
}, (request, reply) => {
 // ...
})

fastify.route({
  method: 'GET',
  path: '/two',
  compress: { createGzip: () => zlib.createGzip() },
  decompress: { createGunzip: () => zlib.createGunzip() },
  handler: (request, reply) => {
   // ...
  }
})
```

darkgl0w:
I have started to work on the last part of the proposal with a big overhaul on tests :
[...]

PR #199, opened 2021-11-26
feat: move to `compress` and `decompress` route options and typescript improvements

darkgl0w:
Hello.

Following https://github.com/fastify/fastify-compress/issues/198 proposal, this PR aims to :
- [x] add `compress` and `decompress` route shorthand configuration properties
- [x] improve typescript types and actually test those types
- [x] update the documentation

~~**Note:** some of those changes are breaking changes and will require a major bump.~~

#### Checklist

- [x] run `npm run test` ~~and `npm run benchmark`~~
- [x] tests ~~and/or benchmarks~~ are included


index.d.ts
```diff
@@ -1,31 +1,64 @@
...
-  requestEncodings?: Array<EncodingToken> 
-  forceRequestEncoding?: EncodingToken
-  onUnsupportedRequestEncoding?: (encoding: string, request: FastifyRequest, reply: FastifyReply) => Error | undefined | null
-  onInvalidRequestPayload?: (encoding: string, request: FastifyRequest, error: Error) => Error | undefined | null
+  export interface RouteOptions extends FastifyCompressRouteOptions {}
 }
 
-declare const fastifyCompress: FastifyPlugin<FastifyCompressOptions>
-export default fastifyCompress;
+declare const fastifyCompress: FastifyPluginCallback<FastifyCompressOptions>
```
darkgl0w, line 63: `FastifyPlugin` is deprecated. Changed to the recommended `FastifyPluginCallback`.
This is a breaking change.

index.d.ts
```diff
@@ -1,31 +1,64 @@
...
+  encodings?: EncodingToken[];
+  forceRequestEncoding?: EncodingToken;
+  global?: boolean;
+  inflateIfDeflated?: boolean;
+  onInvalidRequestPayload?: (encoding: string, request: FastifyRequest, error: Error) => Error | undefined | null;
+  onUnsupportedEncoding?: (encoding: string, request: FastifyRequest, reply: FastifyReply) => string | Buffer | Stream;
+  onUnsupportedRequestEncoding?: (encoding: string, request: FastifyRequest, reply: FastifyReply) => Error | undefined | null;
+  requestEncodings?: EncodingToken[];
+  threshold?: number;
+  zlib?: unknown;
```
darkgl0w, line 21: Previously it was defined as `NodeModule`, but I don't think it's ok, example :
```js
const zlib = require('zlib')

app.register(fastifyCompress, {
  zlib: zlib // this will trigger a typescript argument error if set to NodeModule
})
```
`unknown` is more suited here I think. WDYT ?

mcollina (changes requested):
Could you avoid the breaking change? Just document the new way and copy the old option into the new one.

darkgl0w:
> Could you avoid the breaking change? Just document the new way and copy the old option into the new one.

@mcollina > Yes code wise I think I can avoid the breaking change by applying your suggestion and by throwing a deprecation hint warning with typescript (Typescript wise it should be doable, I will try to research a way of tweaking routes config typescript types).
WDYT in the case where someone makes something like this ? Can I go and consider that the new shorthand provided values take precedence on the old ones ? Is it acceptable to you ?
```js
'use strict'
const app: FastifyInstance = fastify()
app.register(require('fastify-compress'), { global: false })
app.get('/', {
  compress: {
    zlib: {
      createGzip: () => createCustomGzipOne()
    }
  },
  config: {
    compress: {
      zlib: {
        createGzip: () => createCustomGzipTwo() // This one will be the final chosen value. Is it OK ?
      }
    }
  }
}, (req, reply) => {
  // ...
})
```

mcollina:
The outer one takes precendence.

darkgl0w:
Oups sorry I made a mistake in my example, I intended to give precedence to the new `compress` option over the `{ config: compress }` one ?
Have I guessed correctly or is it the other way around ? :p

index.js
```diff
@@ -52,16 +52,21 @@ function compressPlugin (fastify, opts, next) {
 
   // add onSend hook onto each route as needed
   fastify.addHook('onRoute', (routeOptions) => {
-    // Manage compression options
+    // If route config.compress has been set it takes precedence over compress
```
darkgl0w, line 55: @mcollina > I made the change, but I don't know if I misunderstood :rofl: so right now `config.(de)compress` takes prevalence over `(de)compress` options, if it should be the over way around I will make the needed changes :smile:

mcollina (approved):
lgtm, this is ok

index.d.ts
```diff
@@ -1,31 +1,79 @@
...
+}
+
+export interface RouteOptions extends fastify.RouteOptions, FastifyCompressRouteOptions {}
+
+export interface RoutesConfigCompressOptions {
+  /** @deprecated `config.compress` is deprecated, use `compress` shorthand option instead */
+  compress?: RouteCompressOptions | false;
+  /** @deprecated `config.decompress` is deprecated, use `decompress` shorthand option instead */
+  decompress?: RouteDecompressOptions | false;
+}
```
darkgl0w, line 62: Another solution here would be :
```suggestion
export interface RoutesConfigCompressOptions {
  /** @deprecated `config.compress` is deprecated, use `compress` shorthand option instead */
  compress?: never;
  /** @deprecated `config.decompress` is deprecated, use `decompress` shorthand option instead */
  decompress?: never;
}
```
Like this it will trigger a typescript error and the `@deprecated` hint would have high visibility.
But it will be a **breaking change for typescript users**.

mcollina: the deprecation is better

mcollina (approved):
lgtm

Fdawgs (approved):

PR #224, opened 2022-05-04
Update to Fastify v4

mcollina:
Due to the changes introduced in https://github.com/fastify/fastify/pull/2954, we need to change how this plugin is registered from:

```js
  const fastify = Fastify()
  fastify.register(compressPlugin, {
    global: true
  })
  fastify.get('/one', (request, reply) => {
    reply.send(json)
  })
```

to

```js
  const fastify = Fastify()
  await fastify.register(compressPlugin, {
    global: true
  }) // we must use await or after now.
  fastify.get('/one', (request, reply) => {
    reply.send(json)
  })
```

The fundamental problem of the change is that in Fastify v4 routes are now defined synchronously while plugins are defined asynchronously. Therefore requiring the await as there is no library to look for the `onRoute` hook. 

Note that users of nested plugins are not impacted, so the following will still work:

```js
  const fastify = Fastify()
  fastify.register(compressPlugin, {
    global: true
  })
  fastify.register(async (fastify) => {
    fastify.get('/one', (request, reply) => {
      reply.send(json)
    })
  })
```

Are we ok with the change?

cc @metcoder95 @fastify/core @fastify/plugins

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ The plugin automatically decides if a payload should be compressed based on its 
 ### Global hook
 The global compression hook is enabled by default. To disable it, pass the option `{ global: false }`:
 ```javascript
-fastify.register(
-  require('@fastify/compress'),
+await fastify.register(
+  import('@fastify/compress'),
   { global: false }
 )
 ```
@@ -46,8 +46,8 @@ Important note! If you are using `@fastify/compress` plugin together with `@fast
 ### Per Route options
 You can specify different options for compression per route by passing in the `compress` options on the route's configuration.
 ```javascript
-fastify.register(
-  require('@fastify/compress'),
+await fastify.register(
+  import('@fastify/compress'),
   { global: false }
 )
 
@@ -74,21 +74,19 @@ Note that the compress method is configured with either the per route parameters
 the plugin was defined as global.
 
 ```javascript
-const fs = require('fs')
-const fastify = require('fastify')()
+import fs from 'fs'
+import fastify from 'fastify'
 
-fastify.register(require('@fastify/compress'), { global: false })
+const app = fastify()
+await app.register(import('@fastify/compress'), { global: false })
 
-fastify.get('/', (req, reply) => {
+app.get('/', (req, reply) => {
   reply
     .type('text/plain')
     .compress(fs.createReadStream('./package.json'))
 })
 
-fastify.listen(3000, function (err) {
-  if (err) throw err
-  console.log(`server listening on ${fastify.server.address().port}`)
-})
+await app.listen(3000)
 ```
 
 ## Compress Options
@@ -96,16 +94,16 @@ fastify.listen(3000, function (err) {
 ### threshold
 The minimum byte size for a response to be compressed. Defaults to `1024`.
 ```javascript
-fastify.register(
-  require('@fastify/compress'),
+await fastify.register(
+  import('@fastify/compress'),
   { threshold: 2048 }
 )
 ```
 ### customTypes
 [mime-db](https://github.com/jshttp/mime-db) is used to determine if a `content-type` should be compressed. You can compress additional content types via regular expression.
 ```javascript
-fastify.register(
-  require('@fastify/compress'),
+await fastify.register(
+  import('@fastify/compress'),
   { customTypes: /x-protobuf$/ }
 )
 ```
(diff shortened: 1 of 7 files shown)


Eomm:
> The fundamental problem of the change is that in Fastify v4, routes are now defined synchronously while plugins are defined asynchronously. Therefore requiring the await as there is no library to look for the onRoute hook.

I missed that change! 😱 

Now the `onRoute` position is now like the `onRegister` one

```
const fastify = require('fastify')({ logger: true })
fastify.addHook('onRoute', function hook(routeOptions) {
  console.log('works')
})
fastify.get('/', async (request, reply) => {
  return { hello: 'world' }
})
fastify.addHook('onRoute', function hook(routeOptions) {
  console.log('does not work')
})
fastify.ready()
```

> Are we ok with the change?

Are we discussing if the `onRoute` hooks should be async or not?

mcollina:
> Now the onRoute position is now like the onRegister one

Exactly

> Are we discussing if the onRoute hooks should be async or not?

Yes, exactly

Fdawgs:
I may have missed this in previous issues re v4, but are we moving to ES modules?

mcollina:
No. I have changed to `import` because top-level-await is available only on esm. The snippets would be invalid otherwise. I can restore require or just remove it in favor of `fastifyCompress`

README.md
```diff
@@ -46,8 +46,8 @@ Important note! If you are using `@fastify/compress` plugin together with `@fast
 ### Per Route options
 You can specify different options for compression per route by passing in the `compress` options on the route's configuration.
 ```javascript
-fastify.register(
-  require('@fastify/compress'),
+await fastify.register(
+  import('@fastify/compress'),
```
climba03003, line 50: Import here should not work. It may not throw error, but the function execute is wrong.
https://github.com/fastify/avvio/blob/74334bbc0a470aeca41a42d8b30ff5cebf7a3f21/boot.js#L206-L208
```suggestion
  require('@fastify/compress'),
```

climba03003 (dismissed):

jsumners:
I think this is okay. I'm really wondering if we should move to removing the callback api completely, but that's a radical idea.

We need to fully illustrate this breaking change in the release notes.

index.js
```diff
@@ -551,6 +551,6 @@ function createError (code, message, statusCode) {
 }
 
 module.exports = fp(compressPlugin, {
-  fastify: '3.x',
+  fastify: '4.x',
```
delvedor, line 554: I think here we can keep `≥3`, as the plugin code doesn't change. It only changes how plugin users are registering it. But this code *as is* could still work in v3.x
```suggestion
  fastify: '>=3.x',
```

delvedor (changes requested):
LGTM with a nit.

>  I'm really wondering if we should move to removing the callback api completely, but that's a radical idea.

I agree, maybe deprecating them in v4 and completely remove in v5?

README.md
mcollina: It _actually_ works even without this PR, avvio handles  it, check it out.

mcollina:
v4 is already too late. Let's focus on shipping with the current set of features.

README.md
fox1t: Let's start using more import in our docs when registering plugins. It is really a cool feature that avvio can handle for ages.

fox1t:
Maybe it will generate some confusion initially, but having fewer options might be more straightforward to the users in the long run.

Eomm:
> > Now the onRoute position is now like the onRegister one
> 
> Exactly
> 
> > Are we discussing if the onRoute hooks should be async or not?
> 
> Yes, exactly

talking about the `onRoute` hook, if we compare the async vs sync, I see only pros for the `async` version.
Meanwhile, the sync `onRegister` is a technical obligation, it is not clear to me the reasoning to implement the sync `onRoute`

index.js
mcollina: From past experiences, maintaining that is a nightmare. We need separate tests for v3 and v4. I prefer to bump the major so we can avoid any compatibility issues down the line.

climba03003 (approved):
LGTM.

delvedor (approved):

index.js
```diff
@@ -551,6 +551,6 @@ function createError (code, message, statusCode) {
...
   name: 'fastify-compress'
```
climba03003, line 555:
```suggestion
  name: '@fastify/compress'
```

climba03003:
> Do we still need `engines` in package.json, or does that need updating to node 14.x? 

IHO, `engines` fields should probably be removed, it is hard to manage.

> Also the `version` in package.json wasn't updated with the v6.0.0 release?
I think it just forgotten to push the related commit.

mcollina:
Fixed the tag.

metcoder95:
LGTM as well, thanks for bringing me in the loop. Sorry for the headache, now I think I didn't document the changes properly, sorry for that 🙁 

> I think this is okay. I'm really wondering if we should move to removing the callback api completely, but that's a radical idea.
> 
> We need to fully illustrate this breaking change in the release notes.

Agree

PR #230, opened 2022-06-11
integrate isDeflate and isGzip

Uzlopak:
integrate isDeflate and isGzip
remove corresponding packages from dependencies
wrote unit tests for the above functions

#### Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,21 +1,38 @@
 'use strict'
 
+// https://datatracker.ietf.org/doc/html/rfc1950#section-2
 function isDeflate (buffer) {
   return (
     typeof buffer === 'object' &&
     buffer !== null &&
     buffer.length > 1 &&
-    buffer[0] === 0x78 &&
-    (buffer[1] === 1 || buffer[1] === 0x9c || buffer[1] === 0xda)
+    // CMF
+    (
+      buffer[0] === 0x78 ||
+      buffer[0] === 0x68 ||
+      buffer[0] === 0x58 ||
+      buffer[0] === 0x48 ||
+      buffer[0] === 0x38 ||
+      buffer[0] === 0x28 ||
+      buffer[0] === 0x18 ||
+      buffer[0] === 0x08
+    ) &&
+    // The FCHECK value must be such that CMF and FLG, when viewed as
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -1,7 +1,28 @@
 'use strict'
 
+function isDeflate (buffer) {
```
Eomm, line 3: what is the scope of this change?
Why should we copy-paste those modules into this one? (at least credits should be added)

Fdawgs: One less dependency that could be potentially compromised?

kibertoad (approved):

lib/utils.js
Uzlopak: isDeflate is now spec compliant and definetly a complete new product. I benched the performance of isDeflate an it has still the same performance with the replaced one.

I also added references to the corresponding RFCs to explain the magic numbers.

Fdawgs (approved):

Uzlopak:
@kibertoad 
@Fdawgs 

I have to ask you to again review it. On a smoking break I realized that i can do the CMF header check in isDeflate with bit operations. 

before:
```
valid x 223,947,412 ops/sec ±0.70% (90 runs sampled)
invalid x 831,593,213 ops/sec ±0.88% (93 runs sampled)
```

after:
```
valid x 1,027,199,609 ops/sec ±0.74% (89 runs sampled)
invalid x 1,038,269,211 ops/sec ±0.94% (91 runs sampled)
```

```
'use strict'

const { Suite } = require('benchmark')
const isDeflate = require('./lib/utils').isDeflate

const suite = new Suite()

const fakeDeflate = Buffer.from([0x78, 0x9d])
const fakeGzip = Buffer.from(Buffer.from([0x1f, 0x8b, 0x08]))

suite.add('valid', function () {
  isDeflate(fakeDeflate)
}).add('invalid', function () {
  isDeflate(fakeGzip)
})
suite
  .on('cycle', function (event) {
    console.log(String(event.target))
  })
  .on('complete', function () {
    console.log('Fastest is ' + this.filter('fastest').map('name'))
    console.log('Slowest is ' + this.filter('slowest').map('name'))
  })
  .run({ async: true })

```

Fdawgs (approved):

mcollina (approved):
lgtm

PR #256, opened 2023-03-24
Fix default supported content types

andyrichardson:
Hey there, quick fix to fix support for [content-type syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#syntax).

I encountered this issue after using an endpoint that was specifying encoding and compression wasn't being applied.

## Before
Screenshot 2023-03-24 at 17 20 10

## After
Screenshot 2023-03-24 at 17 19 53

### DIFF
--- a/test/global-compress.test.js
+++ b/test/global-compress.test.js
@@ -3069,7 +3069,7 @@ for (const contentType of defaultSupportedContentTypes) {
 
     fastify.get('/', (request, reply) => {
       reply
-        .type('text/xml; charset=utf-8')
+        .type(contentType)
         .send(createReadStream('./package.json'))
     })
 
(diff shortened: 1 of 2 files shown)


Eomm (changes requested):
Thnks for the PR
Could you add some unit test as the screen shows?

Uzlopak:
@andyrichardson 
@Eomm 
Added unit tests.

test/global-compress.test.js
```diff
@@ -3048,3 +3048,71 @@ test('It should return an error when using `reply.compress()` with a missing pay
...
 })
+
+const defaultSupportedContentTypes = [
+  'application/json',
+  'application/json; charset=utf-8',
+  'application/graphql-response+json',
+  'application/graphql-response+json; charset=utf-8',
+  'application/xml',
+  'application/xml; charset=utf-8',
+  'octet-stream',
```
Uzlopak, line 3059: I realize now, that this is wrong. Should be application/octet-stream. And application/octet-stream does not match?!

Uzlopak: Correction. application/octet-stream is also matching. Probably a bug.

Eomm: does #262 supersedes this PR?

Uzlopak: No. It should result in a merge conflict, which should be resolved.

Uzlopak:
As predicted we had a merge conflict. Resolved it. I guess it is ok to merge it.

Uzlopak:
@Eomm PTAL

Eomm (approved):

Uzlopak:
You are free to provide unit tests ;)

PR #273, opened 2023-10-21
perf(index): convert unused capture groups to non-capture groups

Fdawgs:
See https://github.com/fastify/fastify-autoload/pull/330

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -108,7 +108,7 @@ function fastifyCompress (fastify, opts, next) {
   next()
 }
 
-const defaultCompressibleTypes = /^text\/(?!event-stream)|(\+|\/)json(;|$)|(\+|\/)text(;|$)|(\+|\/)xml(;|$)|octet-stream(;|$)/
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -460,7 +460,7 @@ function getEncodingHeader (encodings, request) {
       // consider the no-preference token as gzip for downstream compat
       // and x-gzip as an alias of gzip
       // ref.: [HTTP/1.1 RFC 7230 section 4.2.3](https://datatracker.ietf.org/doc/html/rfc7230#section-4.2.3)
-      .replace(/\*|x-gzip/g, 'gzip')
+      .replace(/\*|x-gzip/gu, 'gzip')
```
Uzlopak, line 463: Should we extract the regexp.in a separate variable to avoid initializing it evert time?

gurgunday: Even though V8 may optimize objects inside functions (RegExp included), I always see an increase in performance after moving them out and using references

So I'd say yes, better to make sure it's not causing any degradations

Fdawgs: Updated in https://github.com/fastify/fastify-compress/pull/273/commits/73cfbad5e41eef17b9b0cb4649493503cc9a3d96

Uzlopak (approved):

PR #278, opened 2024-01-28
perf!: reduce default brotli compression factor to 4, was before 11 (maximum)

kahlstrm:
BREAKING CHANGE: The default brotli quality value currently is 11, which is not meant to be used in an on the fly compression context. Change this to a more reasonable default value of 4.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,7 +117,12 @@ function processCompressParams (opts) {
   }
 
   const params = {
-    global: (typeof opts.global === 'boolean') ? opts.global : true
+    global: (typeof opts.global === 'boolean') ? opts.global : true,
+    /**
+     * Default of 4 as 11 has a heavy impact on performance.
+     * @see {@link https://blog.cloudflare.com/this-is-brotli-from-origin#testing}
+     */
+    brotliOptions: { [zlib.constants.BROTLI_PARAM_QUALITY]: 4, ...opts.brotliOptions }
   }
 
   params.removeContentLengthHeader = typeof opts.removeContentLengthHeader === 'boolean' ? opts.removeContentLengthHeader : true
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ You can tune compression by setting the `brotliOptions` and `zlibOptions` proper
     brotliOptions: {
       params: {
         [zlib.constants.BROTLI_PARAM_MODE]: zlib.constants.BROTLI_MODE_TEXT, // useful for APIs that primarily return text
-        [zlib.constants.BROTLI_PARAM_QUALITY]: 4, // default is 11, max is 11, min is 0
+        [zlib.constants.BROTLI_PARAM_QUALITY]: 4, // default is 4, max is 11, min is 0
       },
     },
     zlibOptions: {


index.js
```diff
@@ -117,7 +117,8 @@ function processCompressParams (opts) {
   }
 
   const params = {
-    global: (typeof opts.global === 'boolean') ? opts.global : true
+    global: (typeof opts.global === 'boolean') ? opts.global : true,
+    brotliOptions: { [zlib.constants.BROTLI_PARAM_QUALITY]: 4, ...opts.brotliOptions },
```
Uzlopak, line 121:
```suggestion
    brotliOptions: { [zlib.constants.BROTLI_PARAM_QUALITY]: 4, ...opts.brotliOptions }
```

gurgunday:
This looks like a pretty good change

Choosing anything higher than 6 has significant impact on the performance for little benefit 

Ref: https://github.com/dotnet/runtime/issues/26097#issuecomment-388434527

kahlstrm:
> This looks like a pretty good change
> 
> Choosing anything higher than 6 has significant impact on the performance for little benefit
> 
> Ref: [dotnet/runtime#26097 (comment)](https://github.com/dotnet/runtime/issues/26097#issuecomment-388434527)

In this case 4 was chosen as it's apparently a default used by Cloudflare. [Related](https://blog.cloudflare.com/results-experimenting-brotli). Here level 5 provides some advantages at files above 64KB so that would also be an alternative.

Uzlopak:
Yeah, makes sense. I also was avoiding brotli for the same reason, and preferred gzip as it has higher throughput. But maybe with brotli compression factor of 4 this drawback does not exist anymore.

Uzlopak:
what does the title mean? sensible brotli value? Maybe some more distinctive title is good, so that reading the release notes it can be determined what changed.

`perf: reduce default brotli compression factor to 4, was before 11 (maximum)`

gurgunday (approved):
Will lgtm after the title change :)

Uzlopak (approved):
LGTM

Fdawgs (changes requested):
Maybe put an inline comment referencing why we've set the default to 4?

Fdawgs (approved):

SimenB:
I'd argue changing the default is a breaking change, and not a "performance improvement". I agree it's a better default, but the behaviour shouldn't change without the consumer having the chance to decide if they value throughput over compression

climba03003:
> I'd argue changing the default is a breaking change

Yes, it is.

> the behaviour shouldn't change without the consumer having the chance to decide if they value throughput over compression

@SimenB Maybe you overlook the code, it still allow to change the value in `brotliOptions`. The predefined one is before the spread operator which means it is non-configurable.

Uzlopak:
As it is breaking the api contract i also concur , that it is a semver-major.

mcollina:
> I'd argue changing the default is a breaking change, and not a "performance improvement".

I'm ok in bumping a major for this.

---

>  I agree it's a better default, but the behaviour shouldn't change without the consumer having the chance to decide if they value throughput over compression

Is this implemented already? https://github.com/fastify/fastify-compress?tab=readme-ov-file#brotlioptions-and-zliboptions

Uzlopak:
@mcollina 

Yes:
https://github.com/fastify/fastify-compress/blob/9e45ec99935d67e17a05ecfd386bd405f6bdb87f/index.js#L135

SimenB:
Yes, the user can set it (IIRC we did lower it at my previous workplace) - I'm just pointing out changing the default (even if configurable) is breaking 🙂 When I said "user should have the chance to decide" I merely meant a major version will "force" the user to make a conscious choice, rather than a `perf` version, which is usually a patch release which I assume most people don't comb through changelogs for

Uzlopak:
@SimenB 
And to what value did you reduce it?

gurgunday:
8 seems too much 😅, but I fully get that it is much better than the current default

64kb is significant for text data, I don't believe the average size of an HTML page or a CSS file is up there

I think 4 is a good default

kahlstrm:
> 8 seems too much 😅, but I fully get that it is much better than the current default
> 
> 64kb is significant for text data, I don't believe the average size of an HTML page or a CSS file is up there
> 
> I think 4 is a good default

Note that in the snippet zlib is set at 8, which is comparable to brotli quality 4 IIRC. The brotli level inthe snippet is below to level 5.

PR #284, opened 2024-03-26
fix: default brotliOptions does not applied to createBrotliCompress and createBrotliDecompress

stanleyxu2005:
## The problem
I encounted a remarkable performance slow down after migrating from Koa. 
Although the doc mentions [fastify/compress](https://github.com/fastify/fastify-compress/blob/master/README.md?plain=1#L172) already sets the compression level to 4, but it takes too long to compress (3MB for 5 sec) a json. I guess it is configured unexpectedly by default. 

After some investigation, it could be related to the previous merge request #278. 
Unless I explicitly set the compression level to `4`, I will get `11` (BROTLI_DEFAULT_QUALITY). 

## The root cause
The compression level is a member of `brotliOptions.params` not `brotliOptions`. https://nodejs.org/api/zlib.html#class-brotlioptions

And unfortunately the global default option is overwritten by custom option afterwards https://github.com/fastify/fastify-compress/blob/master/index.js#L129

## The change
1. Proposed a fix with additional test cases included. 
2. Added a new test case to ensure with default options, response will be compressed with level `4`. 
3. Changed the existing test case to speicfy the compression level to `8` and check if it is compressed expectedly.

Loop in author of previous commit @kahlstrm as well.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,6 +109,13 @@ function fastifyCompress (fastify, opts, next) {
 }
 
 const defaultCompressibleTypes = /^text\/(?!event-stream)|(?:\+|\/)json(?:;|$)|(?:\+|\/)text(?:;|$)|(?:\+|\/)xml(?:;|$)|octet-stream(?:;|$)/u
+const recommendedDefaultBrotliOptions = {
+  params: {
+    // Default of 4 as 11 has a heavy impact on performance.
+    // https://blog.cloudflare.com/this-is-brotli-from-origin#testing
+    [zlib.constants.BROTLI_PARAM_QUALITY]: 4
+  }
+}
 
 function processCompressParams (opts) {
   /* istanbul ignore next */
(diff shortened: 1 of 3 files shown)


gurgunday (approved):
lgtm

gurgunday:
There is a type error... can you take a look?

stanleyxu2005:
> There is a type error... can you take a look?
I tried to understand the error, but have no clue why it is failed. 
It expects to throw an type error for `inject()`. I dont see the callback is reachable. Strangely, I checkout the master branch it shows the same error, but the pipeline for previous commits on GitHub are successful.

stanleyxu2005:
After some more code reading, I made a follow-up change. The compress level will be set to 4 for the global option only. If route has its own compression options, it will be merged with global options anway. No need to double overwrite.

mcollina (approved):
Lgtm. but for some reasons our types tests are failing.

(This can’t land otherwise, and it seems not independent).

gurgunday:
Can you try this?

```js
appWithoutGlobal.inject(
  {
    method: 'GET',
    path: '/throw-a-ts-arg-error',
    headers: {
      'accept-encoding': 'gzip'
    }
  },
  (err) => {
    expectType<Error | undefined>(err) // add undefined
  }
)
```

stanleyxu2005:
@gurgunday @mcollina by changing `expectType<Error>(err)` to `expectType<Error | undefined>(err)`, all test errors are gone at my local env. Committed a new change, and hope all checks will be passed soon. :D

gurgunday:
Yeah seems like ts behavior changed

I will release a new version of this package

stanleyxu2005:
> Yeah seems like ts behavior changed
> 
> I will release a new version of this package

Indeed. It's an interesting test error. 

Thank you both for reviewing the code change. 

This project is quite well coded and documented. I'm glad that I contributed some bugfix for it.👍

gurgunday (approved):
lgtm

PR #290, opened 2024-03-29
feat: replace `into-stream` to `Readable.from`

### LINKED ISSUE #228: [Fastify-compress] Not working with aws api gateway proxy+
### Prerequisites

### Fastify version

4.0.0

### Plugin version

6.0.0

### Node.js version

16.x

### Operating system

Linux

lambda

### Description

When I register this pluggin I got no response data in browser (disable the plugin, it works).

I got this error report  ERR_CONTENT_DECODING_FAILED.
(In firefix, the message says: Content Encoding Error).

### Steps to Reproduce

* Simple aws api gateway deploy with proxy++

lambda.js

```
import awsLambdaFastify from '@fastify/aws-lambda';
import app from './main.js';

const proxy = awsLambdaFastify(app)
const handler = (event, context, callback) => {
  context.callbackWaitsForEmptyEventLoop = false;
  return proxy(event, context, callback)
};

export {
  handler
}
```

app.js ("@fastify/compress": "^6.0.0",)
```
import mCompress from '@fastify/rate-limit'
 fastify.register(mCompress, {
     global: true,
     encodings: ['gzip', 'deflate']
 })
```

(Even I removed encodings, still not working).
(I run in localhost wihout any problems)

### Expected Behavior

_No response_

### LINKED ISSUE #288: Unicode emoji converted to broken chars in certain extreme long string responses
### Prerequisites

### Fastify version

4.26.2

### Plugin version

7.0.2

### Node.js version

18.x

### Operating system

Windows

### Description

Download a 200kb yaml plain text file with `fastify/compress`. Certain lines of the yaml contains illegal chars. 

image

As you can see, on the left side (without compress), icons in the small green box are same. 
But on the right side, one of the icon becomes 2 strange question marks.

The content encoding is `br` (I tried with `gzip`, same error)

image

Any thoughts, what can be the root cause? 

### Steps to Reproduce

```js
const fastify = new Fastify({})
fastify.register(fastifyCompress)

fastify.get('/subscription', async (req, reply) => {
  const subscription = await handler.downloadClashSubscriptionAsync(config)

  // await require('fs').promises
  //   .writeFile('/out.txt', subscription, 'utf-8')  // <--- the file content is okay

  return reply.send(subscription)  // <--- the file content is corrupt
})
```

### Expected Behavior

Dont expect the two question marks appear in the content.

climba03003:
Fixes #228 
Closes #288 

#### Checklist

- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)


lib/utils.js
```diff
@@ -35,4 +35,12 @@ function isStream (stream) {
   return stream !== null && typeof stream === 'object' && typeof stream.pipe === 'function'
 }
 
-module.exports = { isGzip, isDeflate, isStream }
+// It is a helper used to provide a async iteratable for
+// Readable.from
+async function *intoAsyncIterator (payload) {
+  // string | Buffer
+  yield payload
```
climba03003, line 42: Should be support more types here as `into-stream`?

package.json
```diff
@@ -8,12 +8,13 @@
...
     "@fastify/accept-negotiator": "^1.1.0",
     "fastify-plugin": "^4.5.0",
-    "into-stream": "^6.0.0",
     "mime-db": "^1.52.0",
     "minipass": "^7.0.2",
     "peek-stream": "^1.1.3",
     "pump": "^3.0.0",
-    "pumpify": "^2.0.1"
+    "pumpify": "^2.0.1",
+    "readable-stream": "^4.5.2",
```
gurgunday, line 16: Why do you want to add the dep if it works on all release lines?

package.json
```diff
@@ -8,12 +8,13 @@
...
+    "undici": "^5.28.3"
```
stanleyxu2005, line 17: `undic` is a dev dependency, I'm not mistaken

lib/utils.js
```diff
@@ -35,4 +35,42 @@ function isStream (stream) {
   return stream !== null && typeof stream === 'object' && typeof stream.pipe === 'function'
 }
 
-module.exports = { isGzip, isDeflate, isStream }
+// It is a helper used to provide a async iteratable for
+// Readable.from
+async function * intoAsyncIterator (payload) {
```
climba03003, line 40: I though we should just clone `into-stream` including `License` inside this repository.
Because the logic here is very similar to `into-stream@6` but a bit simpler.

package.json
climba03003: `readable-stream` provides a stable copy of `stream` that prevent strange behavior across different node version.

lib/utils.js
climba03003: cc @mcollina

lib/utils.js
```diff
@@ -35,4 +35,42 @@ function isStream (stream) {
...
+
+  if (
+    (
+    // ArrayBuffer
+      payload instanceof ArrayBuffer ||
+    // NodeJS.TypedArray
+    ArrayBuffer.isView(payload)
+    ) &&
+    // Exclude Buffer to prevent double cast
+    !isBuffer
```
Uzlopak, line 51: first check if it is a buffer then do all the ArrayBuffer stuff.

lib/utils.js
```diff
@@ -35,4 +35,42 @@ function isStream (stream) {
...
+  ) {
+    payload = Buffer.from(payload)
+  }
+
+  // Iterator
+  if (typeof payload === 'object' && typeof payload[Symbol.iterator] === 'function' && !isBuffer) {
```
Uzlopak, line 57: Mayba add fast path if Buffer...

lib/utils.js
```diff
@@ -35,4 +35,42 @@ function isStream (stream) {
...
+    for (const chunk of payload) {
+      yield chunk
+    }
+    return
+  }
+
+  // Async Iterator
+  if (typeof payload === 'object' && typeof payload[Symbol.asyncIterator] === 'function' && !isBuffer) {
```
Uzlopak, line 65: Can asyncIterator be something else than a function?

lib/utils.js
Uzlopak: Imho the logic is trivial. So no licensing needed.

lib/utils.js
climba03003: No, the symbol should always be function. So, it can be loop.

lib/utils.js
climba03003: Having dinner right now, I will update later.

Uzlopak: Enjoy ;)

lib/utils.js
Uzlopak: Can we do something like Symbol.asyncIterator in payload? Should be faster than typeof.

lib/utils.js
climba03003: It doesn't provides any goods but one more branch and `===` check.

lib/utils.js
```diff
@@ -41,28 +41,28 @@ async function * intoAsyncIterator (payload) {
...
+    )
   ) {
     payload = Buffer.from(payload)
   }
 
+  const isObject = typeof payload === 'object'
+
   // Iterator
-  if (typeof payload === 'object' && typeof payload[Symbol.iterator] === 'function' && !isBuffer) {
+  if (!isBuffer && isObject && Symbol.iterator in payload) {
```
stanleyxu2005, line 57: Would you combine sync/async iterators in one block?

```js
if (!isBuffer && isObject) {
  // Iterator
  if (Symbol.iterator in payload) {
    for (const chunk of payload) {
      yield chunk
    }
    return
  }

  // Async Iterator
  if (Symbol.asyncIterator in payload) {
    for await (const chunk of payload) {
      yield chunk
    }
    return
  }
}
```

lib/utils.js
```diff
@@ -35,4 +35,43 @@ function isStream (stream) {
...
+  if (!isBuffer && isObject && Symbol.asyncIterator in payload) {
+    for await (const chunk of payload) {
+      yield chunk
+    }
+    return
+  }
+
+  // string | Buffer
+  yield payload
+}
```
Uzlopak, line 75:
```suggestion
/**
 * Provide a async iteratable for Readable.from
 */
async function * intoAsyncIterator (payload) {
  if (typeof payload === 'object') {
    if (Buffer.isBuffer(payload)) {
      yield payload
      return
    }
  
    if (
      // ArrayBuffer
      payload instanceof ArrayBuffer ||
      // NodeJS.TypedArray
      ArrayBuffer.isView(payload)
    ) {
      yield Buffer.from(payload)
      return
    }

    // Iterator
    if (Symbol.iterator in payload) {
      for (const chunk of payload) {
        yield chunk
      }
      return
    }

    // Async Iterator
    if (Symbol.asyncIterator in payload) {
      for await (const chunk of payload) {
        yield chunk
      }
      return
    }
  }

  // string
  yield payload
}
```

Uzlopak (approved):
LGTM

index.js
```diff
@@ -509,7 +509,7 @@ function maybeUnzip (payload, serialize) {
   // handle case where serialize doesn't return a string or Buffer
   if (!Buffer.isBuffer(buf)) return result
   if (isCompressed(buf) === 0) return result
-  return intoStream(result)
+  return Readable.from(intoAsyncIterator(result))
```
stanleyxu2005, line 512: You called `Readable.from(intoAsyncIterator(result))` 3 times. 
So I'm wondering if it makes sense to export `intoReadable(result)` (or you decide a better name) instead of `intoAsyncIterator`

But you might need to revise the unittests.

climba03003: v8 is good at optimizing the function that always receives the same input. So, instead of two functions to receive different input. One is enough.

gurgunday (approved):
Looks good to me

package.json
```diff
@@ -8,12 +8,12 @@
...
     "@fastify/accept-negotiator": "^1.1.0",
     "fastify-plugin": "^4.5.0",
-    "into-stream": "^6.0.0",
     "mime-db": "^1.52.0",
     "minipass": "^7.0.2",
     "peek-stream": "^1.1.3",
     "pump": "^3.0.0",
-    "pumpify": "^2.0.1"
+    "pumpify": "^2.0.1",
+    "readable-stream": "^4.5.2"
```
gurgunday, line 16: Not blocking but I just don't see the reason for adding _readable-stream_ **if the test suit is passing without it**, I remember PRs that removed that package being merged as well

Also I get the stability argument but with the same point of view one could say that any part of node's API can change in a breaking manner between major versions, no? Then why not install every single library as a package?

Maybe _readable-stream_ is special, so just asking to learn

Uzlopak: We can ensure that bugs in streams not solved in older node versions, like in node14, are not affecting our implementation.

stanleyxu2005:
@climba03003 I checked out this PR. New test case is failing at my local env. Not sure, why it is passed successfully in pipeline.

PR #363, opened 2025-06-19
feat: add support for compressing Response objects from Fetch API

### LINKED ISSUE #309: Unsupported compression of streams
### Prerequisites

### Fastify version

4.28.1

### Plugin version

7.0.3

### Node.js version

20.15.1

### Operating system

macOS

14.6.1

### Description

On an async function, I return a stream for the body.
If the user doesn't request compression, then it works as expected.
But if he asks for compression, then he gets a 500 and the following error:

```
{"statusCode":500,"code":"ERR_INVALID_ARG_TYPE","error":"Internal Server Error","message":"The \"string\" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of ReadableStream"}
```

### Minimal reproducible example

Configure a basic project and install required modules:

```sh
mkdir mre && cd mre # Create a new directory
npm init -y # Create a new package.json file
npm pkg set type="module" # Make sure type=module in package.json
npm i fastify @fastify/compress # Install required dependencies for the repro
```

**index.js:**

```js
import fastify from 'fastify'
import fastifyCompress from '@fastify/compress'

const app = fastify()
await app.register(fastifyCompress)

const apiUrl = 'https://jsonplaceholder.typicode.com/posts'

// Here it's working fine, but we lose the streaming capability
app.get('/', async (_req, reply) => {
  const res = await fetch(apiUrl);
  const data = await res.text(); // We need to wait for the whole response to be downloaded

  reply
    .type('application/json')
    .send(data)
  return reply
})

// Here it's broken if we ask compression, because we can't send a ReadableStream
app.get('/broken', async (_req, reply) => {
  const res = await fetch(apiUrl);
  const data = res.body; // We get a ReadableStream (see: https://developer.mozilla.org/en-US/docs/Web/API/Request/body)

  reply
    .type('application/json')
    .send(data)
  return reply
})

await app.listen({ port: 3000 })
```

Example of requests:

```sh
# Compression not asked by the client
curl -v http://localhost:3000/ # OK
curl -v http://localhost:3000/broken # OK (even if the path is called "broken")

# Compression asked by the client
curl -v --compressed http://localhost:3000/ # OK
curl -v --compressed http://localhost:3000/broken # Not OK
```

All requests are working as expected, except the last one.

Here is the output I'm getting:

```sh
> curl -v --compressed http://localhost:3000/broken

* Host localhost:3000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:3000...
* Connected to localhost (::1) port 3000
> GET /broken HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/8.7.1
> Accept: */*
> Accept-Encoding: deflate, gzip
>
* Request completely sent off
< HTTP/1.1 500 Internal Server Error
< content-type: application/json; charset=utf-8
< content-length: 219
< Date: Wed, 21 Aug 2024 15:13:36 GMT
< Connection: keep-alive
< Keep-Alive: timeout=72
<
* Connection #0 to host localhost left intact
{"statusCode":500,"code":"ERR_INVALID_ARG_TYPE","error":"Internal Server Error","message":"The \"string\" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of ReadableStream"}
```

### Link to code that reproduces the bug

_No response_

### Expected Behavior

I expect that in the reproduction example, the `curl -v --compressed http://localhost:3000/broken` command is returning the expected output to the user, and not an error.

Maybe it's a missing feature, and implementing chunk-encoding for streams would solve this?

Aeolun:
*After I made the other PR's, I realized there was already an issue open for this, and so I've made an attempt to make a more comprehensive fix to the supported types using Claude Code. It all seems proper to me, and both the new and old tests pass, but still, it's mostly LLM work, so discretion is adviced.*

If this is merged, there is no need for #361 

## Summary
  - Adds support for compressing Response objects from the Fetch API
  - Adds support for compressing raw ReadableStream objects (Web Streams API)
  - Implements automatic extraction of body stream from Response objects
  - Updates documentation and TypeScript types
  - I got really nervous about inadvertently breaking something, so I added a set of integration tests that validate we get the expected results with a set of common HTTP request libraries.

  ## Details
  This PR addresses issue #309 by implementing proper handling of Response objects and ReadableStream
   objects in fastify-compress.

  ### Key Changes:

  1. **Extended payload type support**:
     - Response objects (from Fetch API) - extracts the body stream for compression
     - ReadableStream objects (Web Streams API) - converts to Node.js streams for compression
     - Updated `isCompressiblePayload` to detect these new types

  2. **Stream conversion**:
     - Added `convertResponseToStream` helper that extracts the body from Response objects
     - Uses `Readable.fromWeb()` to convert Web Streams to Node.js streams
     - Handles Response objects with null bodies gracefully

  3. **Focused middleware behavior**:
     - The compress middleware only handles compression-related headers (`Content-Encoding`, `Vary`)
     - Does not copy headers or status from Response objects
     - Maintains clear separation of concerns

  4. **Documentation updates**:
     - Added "Supported payload types" section listing all supported types
     - Updated `reply.compress()` examples with Response and ReadableStream usage
     - Clarified that threshold only applies to strings/Buffers (not streams)
     - Updated gotchas section for general unsupported types

  5. **TypeScript support**:
     - Updated type definitions to include `Response | ReadableStream` in compress method

  ## Tests added
  - [x] Added test for Response objects with body streams using global compression
  - [x] Added test for Response objects using reply.compress()
  - [x] Added test for Response objects without body (null body)
  - [x] Added test for raw ReadableStream compression (global)
  - [x] Added test for raw ReadableStream using reply.compress()
  - [x] All existing tests pass without regressions
  - [x] TypeScript tests pass

  ## Breaking Changes
  None - this is a backwards-compatible enhancement.

  Closes #309

  🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ fastify.get('/custom-route', {
 ### `reply.compress`
 This plugin adds a `compress` method to `reply` that compresses a stream or string based on the `accept-encoding` header. If a JS object is passed, it will be stringified to JSON.
 
-> ℹ️ Note: When compressing a Response object, the compress middleware only extracts and compresses the body stream. It will handle compression-related headers (like `Content-Encoding` and `Vary`) but does not copy other headers or status from the Response object - these remain the responsibility of your application or Fastify's built-in handling.
+> ℹ️ Note: When compressing a Response object, the compress middleware will copy headers and status from the Response object, unless they have already been explicitly set on the reply. The middleware will then compress the body stream and handle compression-related headers (like `Content-Encoding` and `Vary`).
 
 The `compress` method uses per-route parameters if configured, otherwise it uses global parameters.
 
--- a/index.js
+++ b/index.js
@@ -285,6 +285,18 @@ function buildRouteCompress (_fastify, params, routeOptions, decorateOnly) {
 
       // Handle Response objects
       if (payload instanceof Response) {
+        // Copy headers from Response object unless already set
+        for (const [key, value] of payload.headers.entries()) {
+          if (!reply.hasHeader(key)) {
+            reply.header(key, value)
+          }
+        }
+        
+        // Set status code if it's still the default 200 and Response has a different status
+        if (reply.statusCode === 200 && payload.status && payload.status !== 200) {
+          reply.code(payload.status)
+        }
+        
         const responseStream = convertResponseToStream(payload)
         if (responseStream) {
           payload = responseStream
@@ -423,6 +435,18 @@ function compress (params) {
     if (typeof payload.pipe !== 'function') {
       // Handle Response objects
       if (payload instanceof Response) {
+        // Copy headers from Response object unless already set
+        for (const [key, value] of payload.headers.entries()) {
+          if (!this.hasHeader(key)) {
+            this.header(key, value)
+          }
+        }
+        
+        // Set status code if it's still the default 200 and Response has a different status
+        if (this.statusCode === 200 && payload.status && payload.status !== 200) {
+          this.code(payload.status)
+        }
+        
         const responseStream = convertResponseToStream(payload)
         if (responseStream) {
           payload = responseStream
--- a/test/integration-compress.test.js
+++ b/test/integration-compress.test.js
@@ -62,9 +62,8 @@ const testCases = [
       return response
     },
     expectedBody: { response: 'object', compressed: true },
-    // Response headers are not automatically copied by fastify-compress
-    contentType: null,
-    // Fastify uses default status 200 when Response is converted to stream
+    // Response headers are now copied by fastify-compress
+    contentType: 'application/json',
     checkStatus: 200
   },
   {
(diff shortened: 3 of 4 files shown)


types/index.d.ts
```diff
@@ -26,7 +26,7 @@ declare module 'fastify' {
   }
 
   interface FastifyReply {
-    compress(input: Stream | Input): void;
+    compress(input: Stream | Input | Response | ReadableStream): void;
```
mcollina, line 29: This needs a type test as well.

jsumners (commented):
Scanning through this, I see many extraneous comments, linting errors, and likely useless tests.

mcollina:
I really wish this was done and taken over.

PR #394, opened 2026-04-30
feat: replace peek-stream with native node:stream

hardik-kaji:
### Why 

  The peek-stream package depends on through2 and xtend — all of which are
  legacy userland stream utilities. In particular, [xtend](https://github.com/Raynos/xtend) is no longer maintained.
  Removing peek-stream reduces the dependency tree, avoids maintenance/security risks
   from unmaintained packages, and aligns with the ongoing effort (#386) to replace
  userland stream helpers with native node:stream APIs.
  
### Summary

  - Removes the peek-stream dependency by replacing it with a native createPeekStream
   implementation using node:stream Duplex
  - Fixes a bug where onpeek received the entire buffered data instead of only the
  first maxBuffer bytes when a single write exceeded maxBuffer
  - Moves createPeekStream to lib/utils.js with JSDoc documentation and comprehensive
   unit tests

###  Changes

  - index.js — Remove inline createPeekStream, import it from lib/utils.js, remove
  unused Duplex import
  - lib/utils.js — Add createPeekStream with proper peek slicing logic and JSDoc
  - test/utils.test.js — Add 6 unit tests covering peek correctness, chunked input,
  short streams, error propagation, empty streams, and transform destination

#### Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { Readable: NodeReadable } = require('node:stream')
+const { Readable: NodeReadable, Transform } = require('node:stream')
 
 // https://datatracker.ietf.org/doc/html/rfc8878#section-3.1.1
 function isZstd (buffer) {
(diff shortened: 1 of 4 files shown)


hardik-kaji:
@mcollina @Fdawgs @delvedor Please review 🙏  Required checks are passing

### Tests
Screenshot 2026-04-30 at 15 51 29

### Typescript checks
Screenshot 2026-04-30 at 15 51 36

fritzfelix:
Thank you for this. Unfortunately those EOL packages lead to quite some headaches within corporate environments, thus it's much easier to use fastify in those environments if it's clean.

lib/utils.js
```diff
@@ -104,4 +104,72 @@ async function * intoAsyncIterator (payload) {
...
+  let buf = Buffer.alloc(0)
+  let dest = null
+
+  return new Duplex({
+    write (chunk, encoding, cb) {
+      if (dest) {
+        return dest.write(chunk, encoding, cb)
+      }
+
+      buf = Buffer.concat([buf, chunk])
```
mcollina: Avoid doing this for each chunk, but keep all of them in an array and do the concat only when the combined length pass the threshold.

lib/utils.js
```diff
@@ -104,4 +104,72 @@ async function * intoAsyncIterator (payload) {
...
+      if (dest) {
+        dest.end(cb)
+        return
+      }
+
+      onpeek(buf, (err, stream) => {
+        if (err) return cb(err)
+        dest = stream
+        dest.on('data', (d) => this.push(d))
+        dest.on('end', () => { this.push(null); cb() })
```
mcollina, line 149: these were already added before

mcollina (changes requested):
This seem expensive perf-wise. Maybe we can drop peek-stream and have a better interface for this?

hardik-kaji:
> This seem expensive perf-wise. Maybe we can drop peek-stream and have a better interface for this?

@mcollina Done! Refactored createPeekStream into createLazyTransform with a simpler, more performant interface:

  - Transform instead of Duplex — uses idiomatic transform/flush instead of manual write/final/read with
  event listener wiring
  - Synchronous choose(data) instead of callback-based swap(err, stream) — the caller just returns the
  desired stream directly, no callback overhead
  - Array-based buffering — chunks are collected in an array and Buffer.concat is called only once when the
   threshold is reached (instead of on every chunk)
  - Default maxBuffer = 10 — since both callers always pass 10, it's now the default

  The callers are now much cleaner:
```
  function zipStream (deflate, encoding) {
    return createLazyTransform(function (data) {
      switch (isCompressed(data)) {
        case 1: return new Minipass()
        case 2: return new Minipass()
      }
      return deflate[encoding]()
    })
  }

lib/utils.js
hardik-kaji: Have a look at this: https://github.com/fastify/fastify-compress/pull/394#issuecomment-4370867571

lib/utils.js
hardik-kaji: Have a look at this: https://github.com/fastify/fastify-compress/pull/394#issuecomment-4370867571

hardik-kaji:
@Tony133 can you review this PR?

mcollina:
CI seems very red

PR #395, opened 2026-05-01
fix: replace peek-stream/readable-stream with native node:stream  fixes empty body on Node.js 24

### LINKED ISSUE #393: @fastify/compress silently returns empty body for all encodings on Node.js 24 (root cause: peek-stream / readable-stream incompatibility)
## Environment

| Package | Version |
|---|---|
| `@fastify/compress` | 8.3.1 |
| `fastify` | 5.x |
| Node.js | 24.x (confirmed 24.0.0 → 24.1.0; also reproduced on 24.15.0-alpine) |

## Symptom

All browser-negotiated encodings (`gzip`, `br`, `deflate`) return an empty body. Responses with no `Accept-Encoding`, or with an encoding the plugin doesn't support (e.g. `zstd` — falls back to identity), return the full body correctly.

\`\`\`
# Empty body — Content-Length: 0
curl -s -H "Accept-Encoding: gzip" http://localhost:3000/api/v1/... | wc -c
# → 0

# Full body — works fine
curl -s http://localhost:3000/api/v1/... | wc -c
# → 1842
\`\`\`

The response headers show `Content-Encoding: gzip` (and the correct compressed `Content-Length`) even though the body is actually empty — so the issue is **silent at the HTTP layer** and only manifests when the client tries to decompress.

## Root cause

The plugin's `zipStream()` function (and `unzipStream()`) uses `peek-stream@1.1.3` to buffer the first few bytes of the response stream and detect already-compressed payloads before deciding whether to compress them:

```js
// index.js line ~560
function zipStream (encoding, zlibOptions) {
  return peek({ newline: false, maxBuffer: 10 }, function (data, swap) {
    if (isCompressed(data)) {
      swap(null, new PassThrough())
    } else {
      swap(null, compress(encoding, zlibOptions))
    }
  })
}
```

`peek-stream@1.1.3` is **unmaintained** (last release: 2019) and internally depends on `duplexify` + the legacy `readable-stream` npm package (a userland polyfill of Node's stream APIs targeting older Node versions). These produce a hybrid stream object whose internal `_writableState` and `_readableState` do not satisfy the structural expectations of Node 24's native stream machinery. When a zlib `Transform` is piped through this hybrid in Node 24, `stream.pipeline` / `.pipe()` finishes the writable side before the readable side has consumed all data — the transform flushes and closes, writing nothing out. The result is a correctly-structured HTTP response with `Content-Encoding: gzip` but a body of 0 bytes.

The same `readable-stream` package is imported directly in `index.js` for `Readable.from()`:

```js
// index.js line ~12 — old npm package, NOT node:stream
const { Readable } = require('readable-stream')
```

This is distinct from `lib/utils.js`, which already correctly uses the native runtime:

```js
// lib/utils.js line 3 — correct
const { Readable: NodeReadable } = require('node:stream')
```

The inconsistency means the core compression path runs through a different (outdated) stream implementation than the utility helpers.

## Related issues

- #355 — "peek-stream causes random issues" — earlier reports of `peek-stream` instability under specific Node versions
- #369 — "Default encoding changes on Node 22.15+" — `zstd` added to Node's native `zlib` module on 22.15, causing `@fastify/compress` to silently switch defaults, breaking clients that don't advertise `zstd` support

Both prior issues point at the same underlying dependency chain. Node 24 is the version where the `peek-stream` / `readable-stream` incompatibility becomes a complete regression rather than an intermittent one.

## Why `peek-stream` is unnecessary

`lib/utils.js` already provides pure Buffer-based magic-byte detection:

```js
function isGzip (buf) { return buf[0] === 0x1f && buf[1] === 0x8b }
function isDeflate (buf) { return buf[0] === 0x78 && (buf[1] === 0x01 || buf[1] === 0x9c || buf[1] === 0xda) }
function isZstd (buf) { return buf[0] === 0x28 && buf[1] === 0xb5 && buf[2] === 0x2f && buf[3] === 0xfd }
```

These functions already have everything needed to detect already-compressed payloads without peeking into a stream at all. The only reason `peek-stream` exists in the pipeline is to get the first chunk of the stream so these checks can run. A native `Transform` can do that without the unmaintained dependency.

## Proposed fix

Two changes in `index.js`:

**1. Replace the legacy `readable-stream` import with the native runtime module:**

```js
// Before
const { Readable } = require('readable-stream')

// After
const { Readable, Transform, PassThrough } = require('node:stream')
```

**2. Rewrite `zipStream()` and `unzipStream()` using a native `Transform` that buffers the first chunk, runs the existing magic-byte checks, then splices in the real stream:**

```js
function zipStream (encoding, zlibOptions) {
  let handled = false
  const out = new PassThrough()

  const transform = new Transform({
    transform (chunk, _enc, callback) {
      if (handled) {
        callback(null, chunk)
        return
      }
      handled = true
      if (isCompressed(chunk)) {
        // Already compressed — pass through unchanged
        out.write(chunk)
        callback()
      } else {
        // Compress — splice in the real zlib stream
        const compressor = compress(encoding, zlibOptions)
        compressor.pipe(out)
        compressor.write(chunk, callback)
        this.on('data', (c) => compressor.write(c))
        this.on('end', () => compressor.end())
      }
    },
    flush (callback) {
      out.end()
      callback()
    }
  })

  transform.pipe(out)
  return out
}
```

> **Note:** The exact Transform wiring above is indicative. A cleaner approach (and the one we've taken in our own codebase as a workaround) is to collect the payload as a Buffer before the hook fires — since Fastify's `onSend` provides string/Buffer payloads for non-streaming responses — run the magic-byte check directly on that Buffer, and only create the compressor if needed. No peek-stream or Transform splicing required at all.

## Workaround (for projects blocked now)

Replace `@fastify/compress` with a custom `onSend` hook that uses Node's native `zlib` directly:

```ts
import { createBrotliCompress, createDeflate, createGzip } from 'zlib';
import fp from 'fastify-plugin';

function compressBuffer(buf, encoding) {
  return new Promise((resolve, reject) => {
    const stream = encoding === 'br' ? createBrotliCompress()
                 : encoding === 'gzip' ? createGzip()
                 : createDeflate();
    const chunks = [];
    stream.on('data', (c) => chunks.push(c));
    stream.on('end', () => resolve(Buffer.concat(chunks)));
    stream.on('error', reject);
    stream.end(buf);
  });
}

export const compressionPlugin = fp(async (fastify) => {
  fastify.addHook('onSend', async (request, reply, payload) => {
    if (reply.hasHeader('content-encoding')) return payload;
    const ae = request.headers['accept-encoding'] ?? '';
    const encoding = ae.includes('br') ? 'br'
                   : ae.includes('gzip') ? 'gzip'
                   : ae.includes('deflate') ? 'deflate'
                   : null;
    if (!encoding) return payload;
    let buf;
    if (Buffer.isBuffer(payload)) buf = payload;
    else if (typeof payload === 'string') buf = Buffer.from(payload, 'utf8');
    else return payload;
    if (buf.length < 1024) return payload;
    const compressed = await compressBuffer(buf, encoding);
    reply.header('content-encoding', encoding);
    reply.removeHeader('content-length');
    return compressed;
  });
}, { fastify: '5.x', name: 'compression' });
```

This bypasses `peek-stream` and `readable-stream` entirely and produces correct output on Node 24.

## Suggested path forward for the package

1. Remove `peek-stream` and `readable-stream` from `package.json` dependencies
2. Replace `require('readable-stream')` with `require('node:stream')` throughout `index.js`
3. Rewrite `zipStream()` / `unzipStream()` using the existing `isGzip()` / `isDeflate()` / `isZstd()` helpers from `lib/utils.js` — the magic-byte infrastructure is already there
4. Add a Node 24 entry to the CI test matrix to prevent regression

Happy to submit a PR for this if that would be helpful.

AbdelatefElshafei:
Closes #393

## Problem

On Node.js 24, all compressed responses silently return an empty body despite
correct `Content-Encoding` headers. Requests without `Accept-Encoding` (identity
path) work fine.

Root cause identified in #393: `peek-stream@1.1.3` (unmaintained since 2019)
depends on `readable-stream` (a userland polyfill). On Node 24 the hybrid stream
objects they produce are structurally incompatible with native stream machinery
the pipeline closes the writable side before any data reaches the readable side.

## Solution

Drop the three affected dependencies and rewrite using native `node:stream`:

- **[zipStream()](cci:1://file:///c:/Users/pc/Documents/assets/fastify-compres/index.js:555:0-593:1)**  native `Transform` that checks the first chunk with the
  existing [isGzip](cci:1://file:///c:/Users/pc/Documents/assets/fastify-compres/lib/utils.js:35:0-47:1)/[isDeflate](cci:1://file:///c:/Users/pc/Documents/assets/fastify-compres/lib/utils.js:18:0-33:1) magic-byte helpers, then either passes through
  (already compressed) or lazily wires in the real zlib compressor
- **[unzipStream()](cci:1://file:///c:/Users/pc/Documents/assets/fastify-compres/index.js:595:0-641:1)**  same pattern, preserving the recursive
  `compose(inflate.gzip/deflate(), unzipStream(...))` decompression chain
- **Imports** — `Readable` and `Transform` sourced from `node:stream`

## Changes

| File | Change |
|---|---|
| [index.js](cci:7://file:///c:/Users/pc/Documents/assets/fastify-compres/index.js:0:0-0:0) | Remove `peek-stream`/`minipass`/`readable-stream` imports; rewrite [zipStream()](cci:1://file:///c:/Users/pc/Documents/assets/fastify-compres/index.js:555:0-593:1) and [unzipStream()](cci:1://file:///c:/Users/pc/Documents/assets/fastify-compres/index.js:595:0-641:1) |
| [package.json](cci:7://file:///c:/Users/pc/Documents/assets/fastify-compres/package.json:0:0-0:0) | Remove `peek-stream`, `readable-stream`, `minipass` from dependencies |
| [test/global-compress.test.js](cci:7://file:///c:/Users/pc/Documents/assets/fastify-compres/test/global-compress.test.js:0:0-0:0) | Add multi-chunk uncompressed stream test + zstd malformed-zlib fallback test |
| [.c8rc](cci:7://file:///c:/Users/pc/Documents/assets/fastify-compres/.c8rc:0:0-0:0) | Exclude [eslint.config.js](cci:7://file:///c:/Users/pc/Documents/assets/fastify-compres/eslint.config.js:0:0-0:0) from coverage |

## Testing
node --test 
 180 pass, 0 fail 
npm run test:coverage 
100% lines / functions / statements / branches

### DIFF
--- a/.c8rc
+++ b/.c8rc
@@ -3,7 +3,8 @@
   "exclude": [
     "test/",
     "coverage/",
-    "types/"
+    "types/",
+    "eslint.config.js"
   ],
   "clean": true,
   "check-coverage": true,
--- a/index.js
+++ b/index.js
@@ -2,14 +2,11 @@
 
 const zlib = require('node:zlib')
 const { inherits, format } = require('node:util')
-const { pipeline, compose } = require('node:stream')
+const { pipeline, compose, Readable, Transform } = require('node:stream')
 
 const fp = require('fastify-plugin')
 const encodingNegotiator = require('@fastify/accept-negotiator')
 const mimedb = require('mime-db')
-const peek = require('peek-stream')
-const { Minipass } = require('minipass')
-const { Readable } = require('readable-stream')
 
 const { isStream, isGzip, isDeflate, intoAsyncIterator, isWebReadableStream, isFetchResponse, webStreamToNodeReadable } = require('./lib/utils')
 
@@ -557,26 +554,90 @@ function maybeUnzip (payload, serialize) {
 }
 
 function zipStream (deflate, encoding) {
-  return peek({ newline: false, maxBuffer: 10 }, function (data, swap) {
-    switch (isCompressed(data)) {
-      case 1: return swap(null, new Minipass())
-      case 2: return swap(null, new Minipass())
+  let compressor = null
+  let isPassthrough = false
+
+  return new Transform({
+    transform (chunk, _enc, callback) {
+      if (isPassthrough) {
+        callback(null, chunk)
+        return
+      }
+
+      if (compressor !== null) {
+        compressor.write(chunk, callback)
+        return
+      }
+
+      switch (isCompressed(chunk)) {
+        case 1:
+        case 2:
+          isPassthrough = true
+          callback(null, chunk)
+          return
+      }
+
+      compressor = deflate[encoding]()
+      compressor.on('data', (c) => this.push(c))
+      compressor.on('error', (err) => this.destroy(err))
+      compressor.write(chunk, callback)
+    },
+    flush (callback) {
+      if (compressor !== null) {
+        compressor.once('end', callback)
+        compressor.end()
+      } else {
+        callback()
+      }
     }
-    return swap(null, deflate[encoding]())
   })
 }
 
 function unzipStream (inflate, maxRecursion) {
   if (!(maxRecursion >= 0)) maxRecursion = 3
-  return peek({ newline: false, maxBuffer: 10 }, function (data, swap) {
-    // This path is never taken, when `maxRecursion` < 0 it is automatically set back to 3
-    /* c8 ignore next */
-    if (maxRecursion < 0) return swap(new Error('Maximum recursion reached'))
-    switch (isCompressed(data)) {
-      case 1: return swap(null, compose(inflate.gzip(), unzipStream(inflate, maxRecursion - 1)))
-      case 2: return swap(null, compose(inflate.deflate(), unzipStream(inflate, maxRecursion - 1)))
+  let decompressor = null
+  let isPassthrough = false
+
+  return new Transform({
+    transform (chunk, _enc, callback) {
+      if (isPassthrough) {
+        callback(null, chunk)
+        return
+      }
+
+      if (decompressor !== null) {
+        decompressor.write(chunk, callback)
+        return
+      }
+
+      switch (isCompressed(chunk)) {
+        case 1: {
+          decompressor = compose(inflate.gzip(), unzipStream(inflate, maxRecursion - 1))
+          decompressor.on('data', (c) => this.push(c))
+          decompressor.on('error', (err) => this.destroy(err))
+          decompressor.write(chunk, callback)
+          return
+        }
+        case 2: {
+          decompressor = compose(inflate.deflate(), unzipStream(inflate, maxRecursion - 1))
+          decompressor.on('data', (c) => this.push(c))
+          decompressor.on('error', (err) => this.destroy(err))
+          decompressor.write(chunk, callback)
+          return
+        }
+      }
+
+      isPassthrough = true
+      callback(null, chunk)
+    },
+    flush (callback) {
+      if (decompressor !== null) {
+        decompressor.once('end', callback)
+        decompressor.end()
+      } else {
+        callback()
+      }
     }
-    return swap(null, new Minipass())
   })
 }
 
--- a/package.json
+++ b/package.json
@@ -8,10 +8,7 @@
   "dependencies": {
     "@fastify/accept-negotiator": "^2.0.0",
     "fastify-plugin": "^5.0.0",
-    "mime-db": "^1.52.0",
-    "minipass": "^7.0.4",
-    "peek-stream": "^1.1.3",
-    "readable-stream": "^4.5.2"
+    "mime-db": "^1.52.0"
   },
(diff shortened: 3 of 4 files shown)


Fdawgs:
@AbdelatefElshafei read rule 10 of contributing.
Also please don't tag people in comments expecting reviews the second you open the PR. Half tempted to close this over that.

Pretty sure this is a bot.

Issue #277, opened 2024-01-17
How to cache the compressed responses?


adaboese:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I assume there is a not insignificant overhead of using compression.

Instead, it would be cheaper to cache (or even store on disk) copies of the compressed content.

Is there a hook or another method I could use to achieve this instead of cache-ing every asset adhoc?

mcollina:
Generically, this is better done at the CDN/reverse proxy level by using the Cache-Control headers at the application level.
What makes caching the compressed data _hard_ for this module is knowing when to serve the cached response.

To know if we can serve the cached response, we need to hash the original version, but to do so, we need to have it fully in memory. If it is a _small payload_, it's likely better to transfer it uncompressed. Larger payloads usually _must_ be sent as stream; otherwise, it would exhaust the memory, but if they are streams, we won't know their full hash before we have read them fully.

gurgunday:
See `precompressed`: https://github.com/fastify/fastify-static?tab=readme-ov-file#precompressed

PR #402, opened 2026-06-13
fix: skip compression for 206 and Content-Range responses

naveentehrpariya:
## Summary

A `206 Partial Content` response carries a `Content-Range` header whose byte offsets describe the **original, unencoded** payload. When `@fastify/compress` compresses such a response, the body shrinks but `Content-Range` still describes the pre-compression byte range — a range client stitching parts together gets corrupted data.

The same class of bug was recently fixed in `hono/compress` (honojs/hono#5023) and is tracked in `expressjs/compression` (#277).

## Changes

**`index.js`** — two new guards in the `noCompress` check:

```js
// don't compress partial content: Content-Range describes unencoded byte offsets
(reply.statusCode === 206) ||
(reply.getHeader('Content-Range') !== undefined) ||
```

**`test/global-compress.test.js`** — new `Range Responses` describe block with two tests:

1. `it should not compress 206 Partial Content responses`
2. `it should not compress responses carrying a Content-Range header`

## Test

```
npm test
# 0 failures — new tests pass, all existing tests pass
```

## Checklist

- [x] Tests added
- [x] All tests pass (`npm test` → 0 failures)

### DIFF
--- a/index.js
+++ b/index.js
@@ -258,6 +258,9 @@ function buildRouteCompress (_fastify, params, routeOptions, decorateOnly) {
     const noCompress =
       // don't compress on x-no-compression header
       (req.headers['x-no-compression'] !== undefined) ||
+      // don't compress partial content: Content-Range describes unencoded byte offsets
+      (reply.statusCode === 206) ||
+      (reply.getHeader('Content-Range') !== undefined) ||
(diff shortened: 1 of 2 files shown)


climba03003 (approved):

mcollina (approved):
lgtm

PR #403, opened 2026-06-14
fix: keep the default response encoding stable across Node.js versions

### LINKED ISSUE #369: Default compression silently changes when running under new Node version
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the regression has not already been reported

### Last working version

8.0.0

### Stopped working in version

8.1.0

### Node.js version

22.15+

### Operating system

Linux

alpine 3.22

### 💥 Regression Report

When building an existing app using fastify-compress with a new version of node (going from 20 to 22 in our case), the default compression changes from `br` to `zstd` . It's quite unexpected that the default behavior should change based on which version of node the app is running on. This causes issues for us, since other parts of our stack do not expect (or indeed recognise) `zstd` compression. 

### Steps to Reproduce

Run any project using `fastify-compress` under node < 22.15 . The default compression is `br`. 
Run any project using `fastify-compress` under node > 22.15 . The default compression is `zstd`. 

### Expected Behavior

Default compression should not change when node version changes

rxits:
**Fixes #369**

### Regression
Since v8.1.0, the default response compression changes based on the Node.js version: under Node.js < 22.15 the default is `br`, but under Node.js ≥ 22.15 (where `zlib.createZstdCompress` exists) the same app defaults to `zstd`. Upgrading Node silently changes the wire format, which breaks consumers that don't expect/recognize `zstd`.

### Cause
`processCompressParams` / `processDecompressParams` build the supported-encodings list and then `unshift('zstd')` to the **front** whenever the runtime supports it:

```js
const supportedEncodings = ['br', 'gzip', 'deflate', 'identity']
if (typeof zlib.createZstdCompress === 'function') {
  supportedEncodings.unshift('zstd')   // zstd becomes the top server preference
}
```

`@fastify/accept-negotiator` uses this order to break ties, so a normal client sending `Accept-Encoding: gzip, deflate, br, zstd` now negotiates `zstd` instead of `br`.

### Fix
Insert `zstd` **before `identity`** rather than at the front, so `br` stays the negotiated default (matching pre-8.1.0 behavior) while `zstd` is still offered when a client explicitly prefers it. `identity` remains last. Applied to both the compress and decompress encoding lists.

### Tests
Added a test asserting that a client accepting `gzip, deflate, br, zstd` gets `br` (not `zstd`) on a zstd-capable runtime. Verified failing before the fix (`expected 'br', actual 'zstd'`) and passing after. Full unit suite: 174/174 pass, lint clean. The existing explicit-`zstd` and wildcard (`*` → gzip) tests are unaffected.

### DIFF
--- a/index.js
+++ b/index.js
@@ -154,7 +154,10 @@ function processCompressParams (opts) {
 
   const supportedEncodings = ['br', 'gzip', 'deflate', 'identity']
   if (typeof zlib.createZstdCompress === 'function') {
-    supportedEncodings.unshift('zstd')
+    // Support zstd when the runtime provides it, but keep it below the
+    // historical encodings so the negotiated default does not change with the
+    // Node.js version (see #369). `identity` must stay last.
+    supportedEncodings.splice(supportedEncodings.indexOf('identity'), 0, 'zstd')
   }
 
   params.encodings = Array.isArray(opts.encodings)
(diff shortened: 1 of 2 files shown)


mcollina:
> Upgrading Node silently changes the wire format, which breaks consumers that don't expect/recognize zstd.

How? Compression is negotiated between the client and the server.

PR #411, opened 2026-07-14
fix: suppress native pipeline premature close errors in onEnd

### LINKED ISSUE #410: onEnd's premature-close guard misses Node's native pipeline error ('Premature close' / ERR_STREAM_PREMATURE_CLOSE) — client aborts still log at error level
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.6.1

### Plugin version

9.0.0

### Node.js version

22.x

### Operating system

Linux

### Description

#382 added a guard in `onEnd` so client disconnects mid-response ("premature close") stop logging at error level. The guard compares the error **message** against the lowercase string:

```js
// index.js (v9.0.0, lines 458–466)
function onEnd (err) {
  // Client disconnection during streaming is expected and handled by Fastify.
  // Do not log "premature close" errors at error level since they are not
  // actual errors - they occur when clients disconnect mid-response.
  // See: https://github.com/fastify/fastify-compress/issues/382
  if (err && err.message !== 'premature close') {
    this.log.error(err)
  }
}
```

`'premature close'` (lowercase) is the message produced by the legacy `pump`/`end-of-stream` packages. But `onEnd` is the callback of Node's **native** `require('node:stream').pipeline` (index.js line 5), and native pipeline produces `ERR_STREAM_PREMATURE_CLOSE` whose message is `'Premature close'` — capital P (`node/lib/internal/errors.js`). The comparison is case-sensitive, so the guard never matches the error it was written to suppress, and every client abort during a compressed streaming response still logs at error level.

Observed in production at fleet scale (hundreds of these per day, all benign client disconnects):

```
Error: Premature close
    at Duplexify.<anonymous> (node:internal/streams/pipeline:422:29)
    at Duplexify.emit (node:events:520:35)
    at Duplexify._destroy (.../node_modules/duplexify/index.js:199:8)
    at .../duplexify/index.js:182:10
    at processTicksAndRejections (node:internal/process/task_queues:85:11)
```

(`duplexify` here is `peek-stream`'s stream from this plugin's `zipStream`; `code: 'ERR_STREAM_PREMATURE_CLOSE'`.)

### Steps to Reproduce

```js
const fastify = require('fastify')({ logger: { level: 'debug' } })
const { Readable } = require('node:stream')
const { randomBytes } = require('node:crypto')

await fastify.register(require('@fastify/compress'), { global: true, encodings: ['gzip'] })

fastify.get('/stream', (req, reply) => {
  const source = new Readable({ read () {} })
  const t = setInterval(() => source.push(randomBytes(65536)), 5)
  source.on('close', () => clearInterval(t))
  reply.header('content-type', 'text/plain')
  return reply.send(source)
})

await fastify.listen({ port: 3000 })
// then: curl --compressed http://localhost:3000/stream and Ctrl-C mid-response
// → "Premature close" is logged at level 50 despite the #382 guard
```

### Expected Behavior

The guard suppresses (or downgrades) the native pipeline variant too. Matching on the error **code** is robust against message casing and future wording changes, while keeping the legacy message check for streams that still route through `pump`/`end-of-stream` (which sets no `code`):

```js
if (err && err.code !== 'ERR_STREAM_PREMATURE_CLOSE' && err.message !== 'premature close') {
  this.log.error(err)
}
```

Happy to send a PR if that direction looks right.

xianjianlf2:
## Checklist
- [x] commit message and code follows the [Code of conduct](https://github.com/fastify/fastify/blob/main/CODE_OF_CONDUCT.md)

#382 added a guard in `onEnd` so client disconnects mid-response stop logging at error level, comparing `err.message` against `'premature close'` (lowercase) — the message produced by the legacy `pump`/`end-of-stream` packages. But `onEnd` is the callback of Node's **native** `stream.pipeline`, which emits `ERR_STREAM_PREMATURE_CLOSE` with the message `'Premature close'` (capital P). The case-sensitive comparison never matched, so every client abort during a compressed streaming response still logged at error level.

This PR matches on the error **code** `ERR_STREAM_PREMATURE_CLOSE`, which is stable across Node.js versions and message-wording changes, while keeping the legacy lowercase message check for streams routed through `pump`/`end-of-stream` (which set no `code`).

The existing `issue-382` regression test matched the lowercase message only, so it could not detect the native variant; it now matches any casing. A new `issue-410` regression test covers both the global-compression and `reply.compress` paths (fails with 10/10 aborts logged before the fix) and verifies real errors are still logged.

Closes #410

### DIFF
--- a/index.js
+++ b/index.js
@@ -462,8 +462,12 @@ function onEnd (err) {
   // Client disconnection during streaming is expected and handled by Fastify.
   // Do not log "premature close" errors at error level since they are not
   // actual errors - they occur when clients disconnect mid-response.
+  // Node's native stream.pipeline emits ERR_STREAM_PREMATURE_CLOSE with the
+  // message 'Premature close', while the legacy pump/end-of-stream packages
+  // emit 'premature close' without a code, so both variants are checked.
   // See: https://github.com/fastify/fastify-compress/issues/382
-  if (err && err.message !== 'premature close') {
+  // See: https://github.com/fastify/fastify-compress/issues/410
+  if (err && err.code !== 'ERR_STREAM_PREMATURE_CLOSE' && err.message !== 'premature close') {
     this.log.error(err)
   }
 }
--- a/test/regression/issue-382.test.js
+++ b/test/regression/issue-382.test.js
@@ -15,7 +15,7 @@ test('should not log "premature close" errors on client disconnect (global compr
       level: 'error',
       stream: new Writable({
         write (chunk, encoding, callback) {
-          if (chunk.toString().includes('premature close')) {
+          if (/premature close/i.test(chunk.toString())) {
             prematureCloseCount++
           }
           callback()
@@ -86,7 +86,7 @@ test('should not log "premature close" errors on client disconnect (reply.compre
       level: 'error',
       stream: new Writable({
         write (chunk, encoding, callback) {
-          if (chunk.toString().includes('premature close')) {
+          if (/premature close/i.test(chunk.toString())) {
             prematureCloseCount++
           }
           callback()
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

PR #418, opened 2026-07-29
perf: compress small buffered payloads synchronously

mcollina:
Two related compression performance changes, plus a benchmark to measure them.

## 1. Apply the recommended brotli quality default regardless of `global`

`processCompressParams()` only applied `recommendedDefaultBrotliOptions` (`BROTLI_PARAM_QUALITY: 4`) when `params.global` was true. Registering with `{ global: false }` — the documented manual mode where routes call `reply.compress(payload)` — left `brotliOptions` undefined, so zlib's default quality of **11** was used instead.

The README documents this default as 4, and the constant's own comment notes that quality 11 "has a heavy impact on performance", so this was simply the wrong default rather than an intentional difference.

Compressing 1 MiB with `Accept-Encoding: br` via `reply.compress()`:

| | before | after |
|---|---|---|
| `{ global: true }` | 9 ms | 9 ms |
| `{ global: false }` | **1206 ms** | **9 ms** |

Merge order is unchanged, so a user-supplied `brotliOptions` still takes precedence.

## 2. Compress small buffered payloads synchronously

Every compressed response currently goes through a stream pipeline. For a payload that is already fully in memory, that means allocating a peek transform and a zlib stream per response, and handing the work to the libuv threadpool in several round trips.

For small payloads that costs more than it saves. This adds a `syncThreshold` option: string and Buffer payloads at or below it are compressed in one call, and everything else — larger payloads, streams, web streams, `Response` bodies — keeps the existing pipeline untouched.

Measured over real sockets, 50 concurrent connections, gzip level 6, ~8:1 payloads:

| payload | streamed | synchronous | change |
|---|---|---|---|
| 1 KiB | 17985 r/s | 38909 r/s | **+116%** |
| 2 KiB | 20779 r/s | 30294 r/s | **+46%** |

There is also a memory effect that is independent of payload size, because the synchronous path allocates no per-response zlib context and releases the source buffer immediately instead of pinning it for as long as the response is in flight. Under the same load, RSS dropped from ~348 MiB to ~166 MiB.

### The default scales with host parallelism

The stream path only pays off when there are spare cores to overlap the threadpool work with, so the crossover point depends on how many are available. `syncThreshold` therefore defaults to:

| cores | 1 | 2 | 3+ |
|---|---|---|---|
| default | 65536 | 8192 | 4096 |

Computed once at registration from `availableParallelism()`, clamped to `[4096, 65536]`. libuv's threadpool defaults to 4 threads, so the effect saturates there and the formula does not rely on `UV_THREADPOOL_SIZE` being tuned.

Since libuv 1.49 (Node.js 22.12) `availableParallelism()` accounts for a cgroup CPU quota, so a container given whole cores reports those rather than the host's and gets an appropriate default on its own. Two cases still read as a wide host and land on the smallest default: Node.js before 22.12, which does not consult the quota at all, and a quota below one full core such as Kubernetes `cpu: 500m`, which floors to zero and falls back to the host count. Both err towards the small end, which is the safe direction — settling too low only forgoes some of the win, while settling too high costs considerably more than it gains — and such deployments should set `syncThreshold` explicitly. Quotas are floored rather than rounded, so `cpu: 2500m` reports two cores. This is documented in the README.

Setting `syncThreshold: 0` disables the path entirely and restores the current behaviour exactly.

### Behaviour change

Compressed responses at or below the threshold now carry an accurate `Content-Length` instead of being chunked. The compressed size is known on this path, so there is no stale value to remove — `removeContentLengthHeader` exists because the streaming path cannot know it. When `removeContentLengthHeader: false` and the reply already carries a `Content-Length`, the synchronous path is skipped so the caller-supplied value is preserved byte for byte.

## 3. `npm run benchmark`

Both configurations run against the same build, so no second checkout is needed. The harness is shaped around the ways this measurement goes wrong: the server runs in a forked process, since compressing synchronously occupies the event loop and a co-located client would measure itself; load is driven over real sockets with real concurrency, because `inject()` has neither and cannot observe the stream path overlapping with other cores; and every size is measured repeatedly with medians reported.

It also includes a payload above the threshold as a control. Both configurations stream it, so its result must come out near zero — when it does not, the machine was too noisy and the run is rejected rather than reported. It warns when the load average is high for the same reason.

The numbers above were taken on a shared development machine and the wins at 1–2 KiB are larger than I would quote with confidence; please re-run on an idle host. The memory figures are low-variance and the direction is consistent across every run.

## Notes for reviewers

Two pre-existing issues found while working on this, both left alone as out of scope:

- The `brotliOptions` merge is shallow, so passing any other brotli param still drops the quality default — on `global: true` too. The documented default only holds when no `brotliOptions` are passed at all.
- `Object.assign({}, globalCompressParams, routeParams)` means any route-level `compress` object silently discards globally configured `brotliOptions`. Reproducible on `main`.

`test/regression/issue-350.test.js` is pinned to `syncThreshold: 0`: its ~57 KiB payload sits inside the clamp range, so on a single-core host it would have taken the synchronous path and quietly stopped exercising the streaming premature-close bug it was written for, while still passing.

226 tests pass (188 on `main`), lint and tstyche clean, c8 at 100%. The suite was verified deterministic with `availableParallelism()` stubbed at 1, 2, 3 and 64 cores.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01C75AXneqmj5b9DKKHLvkRt

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,13 +211,21 @@ context per in-flight response — so it is retained in full even at the lowest 
 
 #### Raising it on CPU-constrained deployments
 
-`os.availableParallelism()` reports the host's cores and does not account for a cgroup CPU
-quota. A container limited to a fraction of a CPU on a large host therefore reads as "many
-cores" and gets the smallest default, even though it is precisely the deployment that
-benefits most from compressing synchronously — with little or no spare CPU, there is nothing
-for the streaming path to overlap with.
+Since libuv 1.49 (Node.js 22.12), `os.availableParallelism()` accounts for a cgroup CPU
+quota, so a container limited to whole cores reports those rather than the host's and gets
+an appropriate default on its own. Two cases still read as a wide host and so get the
+smallest default:
 
-If you run under a CPU limit, set `syncThreshold` explicitly:
+| deployment | reported | why |
+| --- | --- | --- |
+| Node.js before 22.12 | host cores | the quota is not consulted at all |
+| a quota below one core, e.g. Kubernetes `cpu: 500m` | host cores | the quota floors to zero and falls back to the host count |
+
+Both are precisely the deployments that benefit most from compressing synchronously: with
+little or no spare CPU, there is nothing for the streaming path to overlap with. Quotas are
+also floored rather than rounded, so `cpu: 2500m` reports two cores.
+
+If either applies to you, set `syncThreshold` explicitly:
 
 ```js
 await fastify.register(
--- a/README.md
+++ b/README.md
@@ -167,6 +167,34 @@ await fastify.register(
   { threshold: 2048 }
 )
 ```
+### syncThreshold
+The maximum byte size for compressing a response synchronously. Defaults to `32768` (32 KiB).
+
+Payloads that are already fully in memory (a `string` or a `Buffer`) and are not larger than
+`syncThreshold` are compressed in one call instead of being pushed through a compression
+stream. This is considerably faster for small responses — a per-response zlib stream costs
+more in setup and scheduling than the compression itself at these sizes — and it uses much
+less memory while the response is in flight, since no zlib context is allocated and the
+source payload is released immediately.
+
+Larger payloads, streams, [web `ReadableStream`s](#supported-payload-types) and fetch
+`Response`s always use the streaming path, so the event loop is never held for long. Raise
+`syncThreshold` to trade more event loop time for more throughput, or set it to `0` to
+compress every response through the streaming path.
+
+```js
+await fastify.register(
+  import('@fastify/compress'),
+  { syncThreshold: 65536 }
+)
+```
+
+Because the compressed size is known upfront, responses compressed synchronously carry an
+accurate `Content-Length` instead of being sent with chunked transfer encoding. Replies that
+already carry a `Content-Length` the plugin has been asked to keep
+(see [removeContentLengthHeader](#manage-content-length-header-removal-with-removecontentlengthheader))
+use the streaming path so the provided value is preserved.
+
 ### customTypes
 [mime-db](https://github.com/jshttp/mime-db) determines if a `content-type` should be compressed. Additional content types can be compressed via regex or a function.
 
@@ -267,6 +295,12 @@ By default, `@fastify/compress` removes the reply `Content-Length` header. Chang
     reply.compress(fs.createReadStream('./file.gz'))
   )
 ```
+
+This option exists because the streaming path cannot know the compressed size upfront, so
+the reply `Content-Length` would otherwise describe the uncompressed payload. Responses
+taking the [synchronous path](#syncthreshold) do know the compressed size and always send an
+accurate `Content-Length`; a reply that already carries a `Content-Length` and sets
+`removeContentLengthHeader` to `false` is streamed instead, so the provided value is kept.
 
 ## Usage - Decompress request payloads
 This plugin adds a `preParsing` hook to decompress the request payload based on the `content-encoding` request header.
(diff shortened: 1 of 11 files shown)


mcollina:
Correction to something I asserted in the description and in a code comment.

I claimed `os.availableParallelism()` ignores a cgroup CPU quota outright. Measured, that is only true up to libuv 1.48. libuv 1.49 consults the quota, and that shipped in Node.js 22.12:

| Node.js | libuv | `--cpus=1.0` reports |
| --- | --- | --- |
| 20.20.2 | 1.46.0 | 16 |
| 22.11.0 | 1.48.0 | 16 |
| 22.12.0 | 1.49.1 | **1** |
| 24.15.0 | 1.51.0 | **1** |

Measured on a 16-core host. So on a current runtime a container given whole cores reports those and gets an appropriate default on its own — which is better than I described, since a 1-CPU container lands on `65536` without configuration.

The guidance still stands, but for two narrower reasons rather than the blanket one:

- Node.js before 22.12 does not consult the quota at all;
- a quota below one full core falls back to the host count, because it floors to zero: `--cpus=0.5` reports 16 while `--cpus=1.0` reports 1. This covers the common Kubernetes `cpu: 500m` shape.

Quotas are floored rather than rounded throughout — `--cpus=1.5` reports 1, `--cpus=2.5` reports 2, `--cpus=3.9` reports 3.

Both remaining cases still read as a wide host and so land on the smallest default, which is the safe direction, and both are covered by setting `syncThreshold` explicitly.

Fixed in f50c64f — comment and README only, no behaviour change. The description has been updated too.

mcollina:
Two more premises behind this PR, verified rather than asserted.

**1. `MAX_USEFUL_PARALLELISM = 4`** — the claim that libuv's threadpool defaults to four threads, so compression cannot scale past four cores and the default should stop shrinking there.

Concurrent `createGzip()` streams, 16 in flight, 512 KiB each, default threadpool, cores restricted with `--cpuset-cpus`:

| cores | 1 | 2 | 3 | 4 | 6 | 8 | 16 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| MiB/s | 19.0 | 35.8 | 51.4 | 64.2 | 68.0 | 66.0 | 64.3 |
| vs 1 core | 1.0x | 1.9x | 2.7x | 3.4x | 3.6x | 3.5x | 3.4x |

Near-linear to four cores, then flat. Cores beyond the fourth add nothing, which is what the cap encodes. The default pool size was confirmed as four separately: with 16 cores available, the default run matches `UV_THREADPOOL_SIZE=4` (63.6 vs 67.9 MiB/s) and is well short of `UV_THREADPOOL_SIZE=8` (97.9 MiB/s).

**2. `Content-Length` on the synchronous path.** The reasoning for setting it — rather than removing it as the streaming path does — is that Fastify recomputes it for Buffer payloads regardless. Confirmed with a plain Fastify app, no plugin, an `onSend` hook returning a 42-byte Buffer:

| what the hook does first | resulting `content-length` |
| --- | --- |
| nothing | 42 |
| `removeHeader("content-length")` | 42 |
| sets it to `999` | 42 |

Fastify overwrites it either way, so the synchronous path cannot avoid emitting an accurate value and `removeContentLengthHeader` has nothing to act on there.

This is also why the `removeContentLengthHeader: false` guard skips the synchronous path when a `Content-Length` is already set: Fastify does not recompute for stream payloads, so routing those responses through the streaming path is the only way a caller-supplied value survives.

Tony133 (approved):
lgtm

PR #420, opened 2026-08-04
fix: preserve status and headers when compressing a Fetch Response

priitkaard:
Compressing a Fetch `Response` drops its status code and headers.

Fastify doesn't unwrap a `Response` before `onSend` - it does that afterwards in `onSendEnd` and
only if the payload is still a `Response`. We swap in the compressed stream before then, so that
never runs and `reply.code()`/`reply.header()` are never called. 

Practically this means `content-type`, `Cache-Control` and `Set-Cookie` disappear and every status
becomes 200.

```js
app.register(compress, { threshold: 0 })
app.get('/', (_req, reply) =>
  reply.send(new Response('hello', { status: 401, headers: { 'set-cookie': 'a=b' } })))
```

```
accept-encoding: gzip      -> 200, no set-cookie
accept-encoding: identity  -> 401, set-cookie: a=b
```

Fix applies the status and headers when we take the body, same as `onSendEnd` would have.

This affects every tRPC v11 app using this plugin, since its Fastify adapter always sends a
`Response`, and those are streams so the threshold never skips them. 
Related issue: trpc/trpc#6628.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,7 @@ const {
   isWebReadableStream,
   isFetchResponse,
   webStreamToNodeReadable,
+  unwrapFetchResponse,
   createPeekTransform
 } = require('./lib/utils')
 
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

climba03003 (approved):

PR #133, opened 2020-10-28
fix: prevent compressing text/event-stream

benjie:
Server-sent events (SSE) are designed to be realtime; by enabling the compression plugin on this content type the latency between writing data to the stream and receiving it on the client can be indefinitely long. This PR prevents compression on event streams, leading to SSE behaving as expected.

This change was provided via the GitHub interface, so I'm hoping that the tests/benchmarks are ran by your CI.

```
> /^text\/|\+json$|\+text$|\+xml$|octet-stream$/.test('text/event-stream')
true
> /^text\/(?!event-stream)|\+json$|\+text$|\+xml$|octet-stream$/.test('text/event-stream')
false
> /^text\/(?!event-stream)|\+json$|\+text$|\+xml$|octet-stream$/.test('text/plain')
true
```

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #421, opened 2026-08-06
fix: unwrap fetch response before noCompress check for SSE support

priitkaard:
onSend decides whether to compress by reading the reply, but when the payload is a Fetch Response fastify only copies its status and headers onto the reply after all onSend hooks have run. The checks therefore see no Content-Type, no Content-Encoding, no Content-Range and a stale status code.

The most visible effect is that the text/event-stream exclusion from #133 never fires for a Response, which is how the tRPC fastify adapter replies. SSE gets compressed and then buffered by zlib until the stream ends, so subscription events all arrive at once. It also hits already-encoded responses (relabelled with the wrong encoding), 206 with Content-Range, and a bodyless Response, which currently 500s on Buffer.byteLength(null).

Regression from #374; #420 fixed the lost headers but kept the ordering.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -329,14 +329,18 @@ function buildRouteCompress (_fastify, params, routeOptions, decorateOnly) {
   }
 
   function onSend (req, reply, payload, next) {
+    if (isFetchResponse(payload)) {
+      payload = unwrapFetchResponse(reply, payload)
+    }
+
     if (payload == null) {
-      return next()
+      return next(null, payload)
     }
 
     const responseEncoding = reply.getHeader('Content-Encoding')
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

Tony133 (approved):

Tony133:
@priitkaard In the future, when making contributions, always create a separate branch for each pull request without using the main branch, as described in the CONTRIBUTING.md document: https://github.com/fastify/.github/blob/main/CONTRIBUTING.md?plain=1#L26 😉

Issue #340, opened 2025-01-02
global options are not taked into account when route options are defined partially


blephy:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.2.0

### Plugin version

8.0.1

### Node.js version

22.12.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

15.1.1

### Description

Given this plugin registered has follow : 

```ts
function onInvalidRequestPayload(
  ...parameters: Parameters<NonNullable<FastifyCompressOptions['onInvalidRequestPayload']>>
) {
  return {
    statusCode: HTTPStatusCodeEnum.BadRequest,
    code: 'Invalid Request Payload',
    name: 'InvalidRequestPayload',
    error: 'Invalid Request Payload',
    message: `Payload is not ${parameters[0]} encoded. ${parameters[2].message}`,
  };
}

function onUnsupportedRequestEncoding(
  ...parameters: Parameters<NonNullable<FastifyCompressOptions['onUnsupportedRequestEncoding']>>
) {
  return {
    statusCode: HTTPStatusCodeEnum.UnsupportedMediaType,
    code: 'Unsupported Media Type',
    name: 'UnsupportedMediaType',
    error: 'Unsupported Media Type',
    message: `${parameters[0]} encoding is not supported`,
  };
}
  
await fastify.register(compressPlugin, {
        global: true,
        onInvalidRequestPayload: onInvalidRequestPayload,
        removeContentLengthHeader: false,
        onUnsupportedRequestEncoding: onUnsupportedRequestEncoding,
      });
```

if a route is registered has follow : 

```ts
fastifyInstance.withTypeProvider<FastifyZodOpenApiTypeProvider>().route({
    method: 'POST',
    url: '/',
    schema: {
      body: BODY,
      response: {
        [HTTPStatusCodeEnum.OK]: RESPONSE,
      },
    },
    decompress: {
      // eslint-disable-next-line @typescript-eslint/explicit-function-return-type, @typescript-eslint/promise-function-async
      onUnsupportedRequestEncoding: function (encoding, request, reply) {
        // we duplicate this here because our preParsing hook is not triggered in this context
        request.monitoringScanEvent =
          MonitoringScanEvent.CreateWithoutTraceId().setName('a');

        return onUnsupportedRequestEncoding(encoding, request, reply);
      },
    },
[...]

PR #397, opened 2026-05-11
fix: merge partial route options with global options

Vinayak1337:
Fixes #340

## What changed

- Merge raw global plugin options with partial route-level `compress` and `decompress` options before processing them once.
- Preserve route option precedence without changing the existing option processors or public API.
- Add regression coverage for inherited handlers and settings, including the `forceRequestEncoding` scenario raised during review.

## Why

The previous `Object.assign({}, globalParams, processParams(routeOptions))` flow allowed defaults and `undefined` fields from the processed route object to overwrite configured global values. Processing one coherent `{ ...globalOptions, ...routeOptions }` object applies the intended route > global > default precedence and follows the simpler approach demonstrated in #398.

## Tests

- `node --test test/routes-compress.test.js test/routes-decompress.test.js`
- `npm run lint`
- `npm test`
- `git diff --check`
- `npm run benchmark --if-present` (executed; the harness rejected the measurements because the unchanged control varied by 12% on a busy machine)

No documentation or type changes are needed because this fixes existing option precedence without changing the public API.

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer Certificate of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11) and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #429, opened 2026-09-02
fix: inherit global options when route options are defined partially (#340)

manuTech0:
# Fix: Inherit Global Options When Route Options Are Defined Partially (#340)

## Problem

Global options (e.g. `onInvalidRequestPayload`, `onUnsupportedRequestEncoding`) registered at the plugin level are ignored when a route defines partial `compress`/`decompress` options.

## Root Cause

`processCompressParams` and `processDecompressParams` always return all fields, including `undefined` for unset values.

During:

```js
Object.assign({}, globalParams, routeParams)
```

the `undefined` values overwrite the global options.

## Fix

Filter out `undefined` values before returning from `processCompressParams` and `processDecompressParams`.

Before:

```js
return params
```

After:

```js
return Object.fromEntries(
  Object.entries(params).filter(([_, value]) => value != undefined)
)
```

This ensures `Object.assign` only applies defined route values, allowing global options to serve as fallbacks.

**Merge priority:** route option > global option > default option

## Checklist

* [ ] Run `npm run test && npm run benchmark --if-present`
* [ ] Tests and/or benchmarks are included
* [ ] Documentation is changed or added
* [ ] Commit message and code follow the [[Developer's Certificate of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11) and the [[Code of Conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,7 +30,7 @@ const defaultSyncThreshold = computeSyncThreshold()
 const InvalidRequestEncodingError = createError('FST_CP_ERR_INVALID_CONTENT_ENCODING', 'Unsupported Content-Encoding: %s', 415)
 const InvalidRequestCompressedPayloadError = createError('FST_CP_ERR_INVALID_CONTENT', 'Could not decompress the request payload using the provided encoding', 400)
 
-function fastifyCompress (fastify, opts, next) {
+function fastifyCompress(fastify, opts, next) {
   const globalCompressParams = processCompressParams(opts)
   const globalDecompressParams = processDecompressParams(opts)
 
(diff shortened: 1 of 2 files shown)


Tony133:
duplicated see PR here: https://github.com/fastify/fastify-compress/pull/397

